PR #63, opened 2022-03-07
Port pulseaudio modules fix to all watches

LecrisUT:
Signed-off-by: Cristian Le <git@lecris.me>


MagneFire:
Thank you for this PR!

Given that this change is applied to all watches, it would, in my opinion, make more sense to just create the file here: https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-multimedia/pulseaudio/pulseaudio-modules-droid_git.bb#L19 instead of having this duplicated for all the watch layers.

LecrisUT:
Ok so there's an issue with my CI triggers, but for now, merge this before meta-asteroid#88

MagneFire:
This change looks good, but can you squash the two commits and remove the `[NoBuild]` tag?

FlorentRevest:
We can squash and rename things from the GitHub UI so I'll do this here

PR #67, opened 2022-03-28
Add support for beluga (OPPO Watch)

MagneFire:
This is a very early WIP on the port for the OPPO Watch that @wannaphong and I have been working on.

I don't think we should merge it yet, but this draft PR let's people know that we have something cooking :smile: 

This currently misses the android system tarball as that keeps changing every now and then. This will be uploaded to Dropbox once ready.

Current state:
- Display works, but fails to wake from sleep (similar to this issue: https://github.com/AsteroidOS/asteroid/issues/191)
- USB works.
- Everything else doesn't. As this is the first Wear 3100 watch we could get to this stage.


scripts/new-watch.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+SRC="$1"
+DST="$2"
```
FlorentRevest, line 3: It would be handy to have a safeguard, if the number of arguments isn't 2 otherwise it's very easy to copy meta- into meta- which is unfortunate. Also this could show a usage info

scripts/new-watch.sh
```diff
@@ -0,0 +1,23 @@
...
+    exit
+fi
+
+cp -r meta-$SRC/ meta-$DST/
+cd meta-$DST/
+FILES=$(find . -name "*$SRC*")
+
+for f in $FILES
+do
+mv "$f" "${f/$SRC/$DST}"
```
FlorentRevest, line 18: Please indent

scripts/new-watch.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo $PWD
+
+if [ -d "meta-$DST/" ]; then
+    echo "Destination layer (meta-$DST) already exists!"
+    exit
+fi
+
+cp -r meta-$SRC/ meta-$DST/
+cd meta-$DST/
```
FlorentRevest, line 13: you can avoid this cd and the cd .. by using find meta-$DST instead of find .

FlorentRevest (approved):

MagneFire:
Okay the latest changes include the ones referenced in: https://github.com/AsteroidOS/meta-smartwatch/pull/74 and https://github.com/AsteroidOS/meta-smartwatch/pull/76.

Additionally, we have Bluetooth and the vibrator working! :tada:

FlorentRevest:
🔥🔥🔥

PR #72, opened 2022-04-04
Narwhal: Fix Bluetooth, sensors and vibration motor

MagneFire:
This PR fixes the most important issues on `narwhal`.
In addition to fixing Bluetooth, sensors and the vibration motor also disable burn-in protection since an LCD doesn't need it.

Also boots the modem. But the microphone doesn't appear to work just yet.


FlorentRevest:
Woop woop! :)

FlorentRevest (approved):

PR #73, opened 2022-04-05
Fix permissions for hands

dodoradio:
It seems that the permission to execute is also needed for all users to have control of the hands. This fixes this. This change is necessary for the PhysicalHands module  (an extension of qml-asteroid to support the physical hands, which I will upstream very soon) to work

### DIFF
--- a/meta-narwhal/recipes-android/android-init/android-init/init.rc
+++ b/meta-narwhal/recipes-android/android-init/android-init/init.rc
@@ -8,7 +8,13 @@ on init
     mkdir /dev/graphics/
     symlink /dev/fb0 /dev/graphics/fb0
     chown system root /sys/class/timed_output/vibrator/enable
-    chmod 666 /sys/devices/sop716/*
+
+    chmod 222 /sys/devices/sop716/motor_move_all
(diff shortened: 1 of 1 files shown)


FlorentRevest (approved):

dodoradio:
So it looks like the numerical value of the permissions is not actually the issue - it seems to be an issue with the way the permissions are being set. I will investigate this further, it was suggested that this may actually be an issue with the * wildcard.

dodoradio:
So after some frustrating tests, I have concluded that the issue was the wildcard I was using in init.rc . I'm still not sure this is the actual issue, there are no docs to confirm or deny this, but it works and doesn't leave and gross security holes, so I'm going to leave it the way it is. Many thanks to @MagneFire for suggesting that this may be the cause of the issue.

MagneFire:
Thanks! This looks good 👍

It does however look like you have some of the lines from the current upstream added in the init.rc (boot DSP and fix vibrator by the looks of it). Can you remove them?

dodoradio:
fixed. Only changes are now those relevant to the hands.

PR #64, opened 2022-03-13
Performance improvements for sturgeon and sparrow.

MagneFire:
These changes improve the ease of porting and has some performance/snappiness improvements.

Please consider merging after https://github.com/AsteroidOS/meta-smartwatch/pull/63 so that I can remove the https://github.com/AsteroidOS/meta-smartwatch/commit/a59c0da19450cf9190b7f42c9ccfd2a9f432804c commit altogether.

The performance numbers compare as follows for `sturgeon`:
- Stock at settings page idle: 30-32 fps
- Stock scrolling the settings page: 24-28 fps
After this patch:
- Stock at settings page idle: 57-59 fps
- Stock scrolling the settings page: 50-55 fps

For `sparrow` the performance differences are also visible:
- Stock: 15 fps
After this patch: 30 - 45 fps.

This works towards fixing the https://github.com/AsteroidOS/asteroid-settings/issues/37 issue.

Big thanks to @eLtMosen for discovering the issue in the first place and helping with the development of these fixes too!

PR #74, opened 2022-04-09
msm8909 based: Performance improvements and black screen fix

MagneFire:
A partial follow up to https://github.com/AsteroidOS/meta-smartwatch/pull/64. We can apply the same optimizations for all our currently supported msm8909 based watches.
One thing to note about this PR is that you may notice the same occurrence of the hybris tarball URL. I think in a future PR we should unify this to a separate recipe (`android-hybris` for example) and use the machine config to define a preferred version to select the hybris tarball to use. I have the following generic tarballs pre-compiled:
- hybris-m-msm8226
- hybris-n-msm8909
- hybris-o-msm8909

Additionally I finally found a workaround/fix for the black screen issue that has been plaguing all msm8909 based watches (exception appears to be `swift`). It's to simply ignore the error. MDP3 is acting in a weird in-between state for some reason.
An overlay is attempted to be 'set' which fails because apparently it has already been set (this is a warning). Then when the overlay is 'play'ed it fails because it hasn't been set, yet the previous function confirms that it has been set.
This fixes: https://github.com/AsteroidOS/asteroid/issues/191


MagneFire:
I've compiled the changes for all affected watches they compile fine now.

FlorentRevest:
This is so cool! :D

FlorentRevest (approved):

Issue #56, opened 2022-02-18
Can't connect mooneye with usb


wannaphong:
I can't connect mooneye with usb. I try to choose adb and developer mode but there don't working.

```
[  418.554988] usb 1-3: new high-speed USB device number 6 using xhci_hcd
[  418.704155] usb 1-3: New USB device found, idVendor=0bb4, idProduct=0c01, bcdDevice= 1.00
[  418.704168] usb 1-3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[  418.704173] usb 1-3: Product: Android
[  418.704177] usb 1-3: Manufacturer: MediaTek
[  418.704181] usb 1-3: SerialNumber: M6600TB1B685
[  536.792287] usb 1-3: USB disconnect, device number 6
[  539.753895] usb 1-3: new high-speed USB device number 7 using xhci_hcd
[  539.903224] usb 1-3: New USB device found, idVendor=0e8d, idProduct=2000, bcdDevice= 1.00
[  539.903228] usb 1-3: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[  539.903229] usb 1-3: Product: MT65xx Preloader
[  539.903231] usb 1-3: Manufacturer: MediaTek
[  540.048654] cdc_acm 1-3:1.0: Zero length descriptor references
[  540.048660] cdc_acm: probe of 1-3:1.0 failed with error -22
[  540.086006] cdc_acm 1-3:1.1: ttyACM0: USB ACM device
[  540.086028] usbcore: registered new interface driver cdc_acm
[  540.086029] cdc_acm: USB Abstract Control Model driver for USB modems and ISDN adapters
[  542.634162] usb 1-3: USB disconnect, device number 7
[  556.045882] usb 1-3: new high-speed USB device number 8 using xhci_hcd
[  556.195101] usb 1-3: New USB device found, idVendor=0e8d, idProduct=2000, bcdDevice= 1.00
[  556.195112] usb 1-3: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[  556.195116] usb 1-3: Product: MT65xx Preloader
[  556.195120] usb 1-3: Manufacturer: MediaTek
[  556.217902] cdc_acm 1-3:1.0: Zero length descriptor references
[  556.217907] cdc_acm: probe of 1-3:1.0 failed with error -22
[  556.298221] cdc_acm 1-3:1.1: ttyACM0: USB ACM device
[...]

PR #75, opened 2022-04-10
usb-moded: Indicate that the system has booted.

MagneFire:
Create a file at /run/systemd/boot-status/init-done to indicate to usb-moded that the system has booted, without this usb-moded would ignore any usb-mode request change and stay in charge mode forever.

This work has not been tested as I don't own a `mooneye` but it's likely to fix https://github.com/AsteroidOS/meta-smartwatch/issues/56 as the issue described there matches the expected behaviour when this patch isn't applied (See commit referenced.)

Ref: https://github.com/AsteroidOS/meta-asteroid/commit/aba49e6e661e522208114854580cc266c54defc1


FlorentRevest:
Hmm, I don't think we really should need this service in the first place, it should be some generic service that detects GadgetFS or FunctionFS.

But that might be outside of this PR...

FlorentRevest (approved):

PR #76, opened 2022-04-10
catfish, ray, sawfish, wren: asteroid-launcher: Remove custom startup service.

MagneFire:
Follows up https://github.com/AsteroidOS/meta-asteroid/pull/86


FlorentRevest:
Why do all these watches provide a custom systemd service for asteroid-launcher ? Can't this be unified ?

FlorentRevest:
> We can do a few other things about this too:
> 
> * Make this just the generic solution for all watches (https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-asteroid/asteroid-launcher/asteroid-launcher/asteroid-launcher.service).

Yes, I believe that would be the right approach

MagneFire:
Alright, the other proposed solution has been implemented in https://github.com/AsteroidOS/meta-asteroid/pull/96.
This means that we can remove all `asteroid-launcher.service` files here.

Only for `wren` there's something special to add. It appeared to be using an additional 5 second delay. However I think it's unlikely that we still need that as the base has moved to Marshmallow and all other fail safes were removed too.

FlorentRevest (approved):

PR #78, opened 2022-04-29
sensorfw: Remove unneeded overrides and adapt to new recipe name.

MagneFire:
sensorfw has been renamed to sensorfw-hybris this is in preparation to the separation of the sensorfw-binder recipe.
With this change the hybris dependency has been added to the sensorfw-hybris package, hence we don't need the override in many places anymore.

This depends on: https://github.com/AsteroidOS/meta-asteroid/pull/98


MagneFire:
Superseded by https://github.com/AsteroidOS/meta-smartwatch/pull/81.

PR #79, opened 2022-04-30
Load WiFi kernel module and adjust bootlogo size

MagneFire:
This automatically loads the kernel module needed to expose a WiFi interface, but it's still disabled/soft-blocked by default (this is good, it means that it doesn't scan for WiFi access points by default).

Additionally, adjust the animated bootlogo so that it matches the screen resolution. This should lead to a clean transition from the bootlogo to the splash animation in the launcher.

Thanks to @eLtMosen for testing these things :smile:


jrtberlin (approved):
Looks alright.

PR #80, opened 2022-04-30
Beluga: Fix boot issue, sensors and tap-to-wake. Improve setup procedure.

MagneFire:
Title says it all :wink:

Basically there was an issue where the kernel wouldn't boot successfully when flashed on the `boot` partition. This needed two patches: disable dm-verity and disable skipping of initramfs.

To fix sensors we only needed to import some binaries from a `beluga` dump and update some of our libraries. See attached PR's below.

Additionally, this has a patch to delay touch reporting to fix tap-to-wake. We already agreed that it's not a nice way to do it, but for now the simplest way until we provide a better patch in mce.

Lastly, a patch is provided to the init script such that the adb shell is enabled when no valid rootfs was detected. This is to allow for installation of the rootfs without booting into WearOS. Booting into WearOS after a factory reset causes for the userdata partition to become encrypted, which makes it impossible to access any pushed file from the AsteroidOS boot image.

This depends on ~https://github.com/AsteroidOS/meta-smartwatch/pull/78~ https://github.com/AsteroidOS/meta-smartwatch/pull/81 and https://github.com/AsteroidOS/meta-asteroid/pull/98.


meta-beluga/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+BOOT_DIR=$1
+
+# Check if the AsteroidOS specific machine configuration file exists.
+# If it doesn't then we know that the userdata partition is mounted but doesn't contain a valid AsteroidOS root.
+if [ ! -e $BOOT_DIR/etc/asteroid/machine.conf ] ; then
+    echo MSG "Push asteroidos.ext4 via ADB" > /run/psplash_fifo
+
+    # Setup for for use with ConfigFS
```
FlorentRevest, line 10: -for

FlorentRevest (approved):

PR #81, opened 2022-05-08
sensorfw: Remove unneeded overrides and adapt to new recipe name.

MagneFire:
This work succeeds https://github.com/AsteroidOS/meta-smartwatch/pull/78

Renames all bbappends to the hal specific recipe plugin and install for most watches except `beluga` as it uses the binder for sensors.


FlorentRevest (approved):
Nice :) Feel free to merge when the other PR is merged

PR #83, opened 2022-05-11
Copy nanaohub bin and build.prop from the original WearOS

LecrisUT:
- This should limit the times the MCU is being flashed, assuming it does not trigger a flash if the firmware versions are the same.
  - [ ] Test by running `/vendor/bin/nanohubapp_cmd download`
- The specific watch can now be queried via `getprop`
- Some files can be removed from the blob `build.prop` and `nanohub.full.*.bin`

### DIFF
--- a/meta-catfish/recipes-android/android/android_catfish-n.bb
+++ b/meta-catfish/recipes-android/android/android_catfish-n.bb
@@ -12,6 +12,7 @@ SRC_URI[hybris.sha256sum] = "626bed275cfe2df2377e709498fc26d58e7883045cd13d4e2a6
 SRC_URI[system.md5sum] = "9082d05811e4ce93efa291a41fd61360"
 SRC_URI[system.sha256sum] = "3c28f9e4579e37c100a24d7d9c693b813b1c90eabf495c5761b855d2f5169708"
 PV = "oreo"
+PR = "r1"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
 INHIBIT_PACKAGE_STRIP = "1"
@@ -19,6 +20,8 @@ COMPATIBLE_MACHINE = "catfish"
 INSANE_SKIP:${PN} = "already-stripped"
 S = "${WORKDIR}"
 B = "${S}"
+
+DEPENDS += "rsync-native"
 
 PROVIDES += "virtual/android-system-image"
 PROVIDES += "virtual/android-headers"
@@ -39,10 +42,12 @@ do_install() {
     rm ${D}${includedir}/android/android-headers.pc
 
     install -d ${D}/system/
-    cp -r system/* ${D}/system/
+    # FIXME: Remove these files from the blob
+    rsync -a --delete \
+      --exclude=vendor/firmware/nanohub.full.*.bin --exclude=build.prop --exclude=vendor/build.prop \
+      system/* ${D}/system/
 
-    cd ${D}
-    ln -s system/vendor vendor
+    ln -s system/vendor ${D}/vendor
 }
 
 # FIXME: QA Issue: Architecture did not match (40 to 164) on /work/dory-oe-linux-gnueabi/android/lollipop-r0/packages-split/android-system/system/vendor/firmware/adsp.b00 [arch]
--- a/meta-catfish/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
+++ b/meta-catfish/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+BOOT_DIR=$1
+
+if [ ! -e $BOOT_DIR/system/build.prop ] ; then
+    mkdir -p /tmp/vendor
+    mkdir -p /tmp/system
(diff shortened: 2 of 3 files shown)


MagneFire:
Thanks for this PR!
Since we have https://github.com/AsteroidOS/meta-smartwatch/pull/85 merged I don't see a good argument for merging this PR since it solves the main issues we had: `nanohub`.

Regarding the other points:
- *getprop*: Really minor, I'd prefer simply setting it to `catfish` for all platforms rather than copying a build.prop from the vendor partition.
  - Ideally we only mount the `vendor` partition in place of using the blob on Dropbox (https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-catfish/recipes-android/android/android_catfish-n.bb#L7). But we still have to figure out how to build the hwcomposer part for this...
- *`nanohub.full.*.bin`* can be removed: To me a more clean solution would be to simply remove it from the tarball on Dropbox directly rather than going through all the efforts of doing that after the fact. I can look into this if you prefer.

LecrisUT:
> Since we have #85 merged I don't see a good argument for merging this PR since it solves the main issues we had: `nanohub`.

Let's not jump the gun yet. We haven't confirmed how it behaves on clean and upgraded WearOS yet. I'll update in a week when my replacement comes.

> `getprop`

If we copy from vendor than we can distinguish between `catfish`, `catshark`, etc. If we have can make the modem functionality become available at runtime, than we can avoid branching the device.

> But we still have to figure out how to build the hwcomposer part for this...

Bit of clarification what you are referring to here? How does `hwcomposer` work? Does it rely on the blob's smelt/sawfish (keep forgetting which) files in /vendor?

> `nanohub.full.*.bin` can be removed

Indeed we should update the tarball with the relevant files left out. But again let's not jump the gun yet, we only have knowledge that a successfully flashed nanohub does not misbehave. We don't know what happens on other nanohubs.

But this also makes me wonder, what/how are we using /vendor, and why mounting the original one fails

MagneFire:
> > Since we have #85 merged I don't see a good argument for merging this PR since it solves the main issues we had: `nanohub`.
> 
> Let's not jump the gun yet. We haven't confirmed how it behaves on clean and upgraded WearOS yet. I'll update in a week when my replacement comes.
> 

Well, we know one thing: sensors remain working if they worked under WearOS. I don't think it's particularly bad if someone needs to install WearOS to fix sensors if that means that it reduces the risk of breaking sensors in the first place.

Fun fact, we already instruct `bass` users to install WearOS when they experience sensor related issues: https://asteroidos.org/install/bass/

> > `getprop`
> 
> If we copy from vendor than we can distinguish between `catfish`, `catshark`, etc. If we have can make the modem functionality become available at runtime, than we can avoid branching the device.
> 

This would never works since we (as you know) rely on the `sawfish` Android blobs (https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-catfish/recipes-android/android/android_catfish-n.bb#L7). For this we should move to mounting the appropriate `/vendor` and `/system` partitions entirely (like what we do for `narwhal`: https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-narwhal/recipes-core/initrdscripts/initramfs-scripts-android/machine.conf#L3). But as mentioned before, we still have issues getting the patched Android 9 base working...

> > But we still have to figure out how to build the hwcomposer part for this...
> 
> Bit of clarification what you are referring to here? How does `hwcomposer` work? Does it rely on the blob's smelt/sawfish (keep forgetting which) files in /vendor?
> 

The `hwcomposer` is what Android uses for graphics rendering. We use a patched version so that we can use it too. The main reason for using this is because it allows us to use the vendor blobs for hardware acceleration.

Some more information about it is located here: https://asteroidos.org/wiki/graphic-stack/

https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-sturgeon/README-system-dir.md gives some insight in how the `hwcomposer` is built. These result in the "generic" tarball located https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-catfish/recipes-android/android/android_catfish-n.bb#L6 (Notice that there are two different tarballs used here.)

> > `nanohub.full.*.bin` can be removed
> 
> Indeed we should update the tarball with the relevant files left out. But again let's not jump the gun yet, we only have knowledge that a successfully flashed nanohub does not misbehave. We don't know what happens on other nanohubs.
> 
> But this also makes me wonder, what/how are we using /vendor, and why mounting the original one fails

This is about `nanohubapp_cmd` right? I'd recommend testing by using the `strace` command (https://release.asteroidos.org/nightlies/ipk/armv7vehf-neon/strace_5.14-r0_armv7vehf-neon.ipk).
It does make somewhat sense as to why it crashes though. All Android binaries (`nanohubapp_cmd` included) are built to look for libraries in `/system` and `/vendor` for libraries. It seems that right now there might be an incompatibility with the libraries in `/vendor` and `/system` which makes sense since the former is from an Android 9 base while the latter is using libraries for an Android 8 base.

LecrisUT:
> This would never works since we (as you know) rely on the `sawfish`

The only thing copied is `build.prop`, which should be pure metadata. Why wouldn't it work?

> The main reason for using this is because it allows us to use the vendor blobs for hardware acceleration.

I'm baffled that hardware acceleration works after mixing different vendor files and hardware.

PR #85, opened 2022-05-16
Removed service updating nanohub from init.rc

dodoradio:
We have had 2 catfish have nanohub bricked so far. Not updating nanohub seems to have no side effects, so it's safer to leave that microcontroller alone until we figure out what's actually causing this issue.
Both of the issues happened on devices that were dual-booted, but we can't be sure that this won't happen on a full flash.


LecrisUT:
Maybe should ping it here in case it gets buried in the chat mentions:
@MagneFire @FlorentRevest @jrtberlin Please get this merged quick and update the image links.

MagneFire (approved):

PR #87, opened 2022-05-28
Disable Always-on-Display on catfish and narwhal by default.

MagneFire:
These watches have some other way to display the time when the screen is off (Additional LCD or physical hands).
Hence we don't need to have Always-on-Display enabled by default. This obviously improves the battery life since the SoC doesn't need to be awakened every minute.

### DIFF
--- a/meta-catfish/recipes-nemomobile/mce/mce/builtin-gconf.values
+++ b/meta-catfish/recipes-nemomobile/mce/mce/builtin-gconf.values
@@ -1,0 +1,2 @@
+/system/osso/dsm/display/display_brightness=5
+/system/osso/dsm/display/use_low_power_mode=false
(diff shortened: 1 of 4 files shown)


FlorentRevest (approved):

FlorentRevest:
This builtin-gconf.values is very neat! :) I wonder if we could use it to replace https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/mce/mce/0001-Keep-screen-on-by-default-on-emulator.patch

MagneFire:
> This builtin-gconf.values is very neat! :) I wonder if we could use it to replace https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/mce/mce/0001-Keep-screen-on-by-default-on-emulator.patch

Indeed, we can replace that patch too. During the porting of the apps to PostmarketOS we also found this out. We could use this https://gitlab.com/postmarketOS/pmaports/-/blob/master/device/main/device-qemu-amd64/mce-display-blanking.conf.

PR #95, opened 2022-06-12
Add support for minnow (Moto 360)

MagneFire:
This adds support for the original Moto 360.

It already seems to be quite functional:
- Display: works, albeit slow due to some GPU issues.
- Touchscreen: Works.
- Vibrator: Works.
- USB: Works fine.
- Sensors: Works, can't really test much here, but the heart rate sensor lights up, the tilt-to-wake sensor shows messages, but doesn't wake the screen. The ambient light sensor also works fine.
- Always-on-Display: Does NOT work, this watch uses the HWComposer API V1.0 which we don't have an implementation for this. Looking through the kernel, it might be a compile time option. Overriding from the hwcomposer plugin might be possible.
- Bluetooth: Does NOT work, but we might be able to use the `btwilink` kernel driver for this as a TI based Bluetooth solution is used.


FlorentRevest:
It looks like the source of the graphical issues is using the fbdev backend (so there's no hw acceleration and everything is software rendered and slow) but that's a good start for a port :) thank you!

FlorentRevest (approved):

MagneFire:
> It looks like the source of the graphical issues is using the fbdev backend (so there's no hw acceleration and everything is software rendered and slow) but that's a good start for a port :) thank you!

The name is what made me think that too. But that's not actually true. The `fbdev` backend (`EGL_PLATFORM`) still uses EGL extensions (hw acceleration).
It just needs to be used on devices that use an older version of the hwcomposer API. (See ref)
I also tried pure software rendering using the `QT_QPA_PLATFORM` (Set to `linuxfb`). This was similarly as slow. But had an additional issue where the `FlatMesh` wasn't rendered.

Ref: https://sailfishos.org/content/uploads/2022/02/SailfishOS-HardwareAdaptationDevelopmentKit-4.3.0.15.pdf See section: 15.10.

Issue #35, opened 2021-05-20
Watch never enters deepsleep.


MagneFire:

Doomsdayrs:
```journalctl -f```
Attempting to run `systemctl suspend`
[2022-03-27-suspend-failure.txt](https://github.com/AsteroidOS/meta-smartwatch/files/8358188/2022-03-27-suspend-failure.txt)

**Comment by Doomsdayrs** [comment]
Random errors that gets pumped out on cycle
```
Mar 27 14:02:31 tetra kernel: bcmpmu_rpc_work, pa_temp 384
```
```
Mar 27 14:02:31 tetra kernel: ADC out of range, raw = 0
```
```
Mar 27 14:02:35 tetra kernel: bcmpmu_read_time: err=0 time=122.2.27.14.2.53
```

**Comment by Doomsdayrs** [comment]
```journalctl --system -f```
[2022-03-27-journalctl.log](https://github.com/AsteroidOS/meta-smartwatch/files/8358199/2022-03-27-journalctl.log)

The MCE is keeping the CPU alive? Why?
```
Mar 27 14:07:00 tetra mce[314]: modules/cpu-keepalive.c: cka_state_set(): keepalive started
```

Pepa77:
hi i don't know how it is possible but last night 15/07/2022 tetra sleeps like not to mention after one days i have 33% battery now Asteroid OS has chance again on this device.

mirh:
There's literally a PR linked to explain how it was possible.

Pepa77:
Yeah sorry I missed it.

Doomsdayrs:
Now to have tetra re-added to the website

eLtMosen:
> Now to have tetra re-added to the website

Since there are still problems with the gpu, @MagneFire proposed to add watches with remaining problems like minnow and tetra to a new "Experimental" section. That way, users would not falsly expect a daily driver ready support level and possibly make a purchase based on those watchs appearing in the general install pages. And developers have an easier time finding it still.

PR #101, opened 2022-07-12
tetra: android-init: Initialize the PMU.

MagneFire:
Finally found a solution to the battery drain issue on `tetra`!
The solution is embarrassingly simple too!
Battery life has improved from barely making it up to 8 hours to an entire day of use with only ~25% drained.

I wasn't sure if it would work with AOD since we don't actually blank the display. But it looks like the PMU handles this perfectly.

I did notice that Bluetooth did become less reliable after this change. Removing the BLE only mode made it more reliable, so we could apply https://github.com/AsteroidOS/meta-smartwatch/commit/85e529c615f87f4cd8c923116b6d37e0059b2f12 here too.

This fixes https://github.com/AsteroidOS/meta-smartwatch/issues/35.

### DIFF
--- a/meta-tetra/recipes-android/android-init/android-init/init.rc
+++ b/meta-tetra/recipes-android/android-init/android-init/init.rc
@@ -3,6 +3,9 @@ on init
     mkdir /dev/graphics/
     symlink /dev/fb0 /dev/graphics/fb0
 
+# PMU init
+    write /sys/module/pwr_mgr/parameters/pm_late_init 1
+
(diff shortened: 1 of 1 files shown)


FlorentRevest (approved):

FlorentRevest:
I'm fine with enabling dual mode if this improves stuff :)

Pepa77:
Thanks for the great report and you're right, 2 full days of battery is incredible 👍

PR #102, opened 2022-07-21
tetra: asteroid-launcher: Enable glyph workaround to fix font rendering issue.

MagneFire:
`tetra` along with many other ports suffer from broken fonts due to a miscompilation in libc.so (from Android).
This results in broken fonts. Qt enables said workaround by default on Android builds but as we are building for Linux this isn't applied.
One can enable it at runtime via a environment variable.

Ref: https://bugreports.qt.io/browse/QTBUG-34984


jrtberlin (approved):

FlorentRevest:
> This is due to a bug in the memmove() implementation for overlapping memory on this particular device, which returns dest + n instead of dest. The compiler in turn recognizes memmove() and uses the return value as an optimization instead of saving the dest pointer on the stack. In the end QVector::prepend() ends up overwriting the full array with the new value instead of just prepending it, which causes memory corruption later on.

:nauseated_face:

Issue #82, opened 2022-05-10
Heart rate sensor no longer works


MagneFire:
It seems that https://github.com/AsteroidOS/meta-sturgeon-hybris/issues/6 has returned...

The hear rate sensor appears to light up and report a heart rate until you restart the heart rate app. At that point `sensorfw` reports an issue where `setInterval` was `not ok`.
My previous findings were that this function is essential for `sturgeon`.
I'll check if downgrading `sensorfw` to an older version fixes this issue again then we can look into creating a fix based on the current version of `sensorfw`.

This issue was reported by @jrtberlin too.

Pepa77:
hello I can confirm that the sensor works only once after restarting the watch, the second time I will not run on the night release of firmware 1.1 for sturgeon 3.6.2022

eLtMosen:
Confirmed on my sturgeon. It works one time (does not matter how long after reboot, the watch  ran the whole day before i tried now). Closing the app and trying again does not light up the sensor.

Pepa77:
hello it would fair on the site settings Asteroid OS for sturgeon heart rate put on red that it broken.

MagneFire:
> hello it would fair on the site settings Asteroid OS for sturgeon heart rate put on red that it broken.

Agreed, I shall update the website until we have a fix.

MagneFire:
I've done some testing and I think the heart rate sensor stopped functioning reliably after this commit: https://github.com/AsteroidOS/meta-asteroid/commit/e3a4c38b5ec3c49d5330546ce0ec232535e0b4d1.

The current plan is to downgrade `sensorfw` prior to this change and test if this makes the heart rate sensor function correctly on current nightlies.

PR #105, opened 2022-08-18
sturgeon: sensorfw: Set default interval for the heart rate sensor.

MagneFire:
The previous method no longer appears to work with recent versions of `sensorfw`.
Instead of setting the interval a single time, set it as a default to prevent it from setting the interval to zero.

This fixes an issue where the heart rate sensor would only start once during a boot.

I initially considered updating the existing patch (https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/sensorfw/files/0002-Add-heart-rate-monitor-sensor-with-hybris-adaptor.patch) but decided against that as this issue is specific to `sturgeon` other watches do not appear to have this issue. But admittedly since the original patch existed for a long time we cannot be entirely sure about that.

This fixes https://github.com/AsteroidOS/meta-smartwatch/issues/82.


beroset (approved):
Tested it today and it works just fine.  Thanks!

FlorentRevest (approved):

FlorentRevest:
Thank you!

PR #109, opened 2022-09-18
catfish: Move to an Android 9 base.

MagneFire:
Like the title says. This is the first port to properly use a patched hybris for Android 9.

Currently in testing and depends on:
- https://github.com/AsteroidOS/meta-asteroid/pull/111
- https://github.com/AsteroidOS/meta-asteroid/pull/112
- https://github.com/AsteroidOS/meta-asteroid/pull/113
- https://github.com/AsteroidOS/meta-asteroid/pull/114

For now it appears to perform slower than the current build, this is because the hybris work doesn't incorporate https://github.com/AsteroidOS/android_hardware_qcom_display/commit/9052e3bac0cc4d1553f55bdf81abfb8aa2066dd8 since it no longer depends on a patched hwcomposer. This means that we need to find a different solution to the problem since at the time it appeared to be resizing of the framebuffer related.

This work adds support for some components that are not supported in current builds:
- Speaker
- GPS
- Maybe more?

I'm not sure when/if we want to merge this since it has the UI performance regression.


FlorentRevest (approved):
The changes LGTM. About the performance issue, I can't test it myself so trust your judgment if/when we should merge this.

MagneFire:
I've done some debugging and it turns out that we cannot solve the performance issue by patching the hwcomposer (https://github.com/AsteroidOS/android_hardware_qcom_display/commit/9052e3bac0cc4d1553f55bdf81abfb8aa2066dd8 and https://github.com/AsteroidOS/android_system_core/commit/e2781ab79687d040d9301e0c5b8c3822e2da6fa3). Said patches no longer effect the performance since the relevant hwcomposer code has been replaced entirely as part of the hwcomposer 2 API (https://android.googlesource.com/platform/hardware/qcom/display/+/d39a514dd0540cf47e121775a77e9ac1b578bdb1%5E%21/).

dodoradio (commented):
I noticed that some files are still labelled `_n`, it is probably worth renaming those to `_p` now so it doesn't get confusing later

MagneFire:
The most recent changes are:
- Use the 'standardized' generic libhybris tarballs.
- Enable more GPU clock frequencies and change the governor.

The latter has improved the performance to a state where I'm happy to merge this PR.

FlorentRevest (approved):
Great work! I just had a nitpick, please feel free to merge it after you fixed it

PR #111, opened 2022-09-25
narwhal: Set timezone to zero to fix a synchronization issue.

MagneFire:
AsteroidOS does not use time zones but WearOS does. This means that WearOS could store the current time including the timezone on physical hands module. Restoring the time on AsteroidOS from the physical hands would cause for an offset in hours as part of the timezone incompatibility. Setting it to zero fixes this until we have proper support for time zones in AsteroidOS.


dodoradio (approved):

FlorentRevest (approved):

FlorentRevest:
Yay :) thanks!

PR #113, opened 2022-10-02
Specify python version in 0002-dtcat-Port-to-Python-3.patch

jrtberlin:
I've seen 3 approaches by distros concerning `python`
- It points to the newest python version
- It points to the latest (now EOL) version of python2
- It doesn't exist

The file is compatible with Python 3 and should therefore declare `python3` explicitly in the shebang.
With this minor patch, we don't depend on either `python2.7` being present or a `python-is-python3` package.


MagneFire (approved):
LGTM! Good catch!

PR #137, opened 2022-12-16
tetra: android: Allow for empty main package

dv1:
This is necessary, otherwise the -dev package cannot be installed in generated SDKs, since the -dev package depends on the main one.


MagneFire (approved):
Great find!

Other platforms might need this as well, but we might need to test this some more for those.

PR #138, opened 2022-12-18
Add support for triggerfish (fossil gen 5  platform)

dodoradio:
Adds support for the Fossil Gen 5 platform. This port was initially cloned from the android 9 catfish port, but now has a lot more in common with beluga. 
Similar to beluga, this watch behaves erratically when attempting to use the `fastboot boot` command, and doesn't actually boot to a kernel this way. 
Similar to ray/firefish, this watch requires removal of the back cover (held on by glue) to access USB. It is definitely worth adding images that show the pinout. I used a 3D printed dock to access USB.
This is still very much a work in progress. 

What works:
- boot to asteroid UI
- touchscreen

What doesn't work:
- rotating input
- bluetooth (I have made basic attempts at getting this working but no success yet)
- audio
- gps
- sensors (probably an easy job, but currently no attempts made)
Other notes: 
- UI is very sluggish and not  very pleasant to use, though this is apparently also the case on beluga. 
- the back cover can be removed with a hair dryer and some persistence.


dodoradio:
just to note that this has only been tested on a `Fossil Gen 5 Carlyle HR` with a dead battery. This is specifically for the `Gen 5` models with DW10xx model codes. There exists also a `Gen 5 LTE` codenamed `gila` and `Gen 5e` codenamed `sole` which won't be supported without additional work, but other Fossil Gen 5 models should be supported, like the `Fossil Q Julianna HR`, `Diesel On Axial`, `Emporio Armani Connected gen 3`, `Michal Kors Bradshaw 2`, `Skagen Falster 3` and various others

meta-triggerfish/recipes-android/android-init/android-init/init.rc
```diff
@@ -0,0 +1,42 @@
...
+    restorecon_recursive /persist
+
+    mkdir /data/
+    mkdir /data/system/
+    mkdir /data/system/gps/
+    mkdir /data/vendor
+    mkdir /data/vendor/sensor
+    mkdir /data/vendor/sensor/nanohub_lock
+    # TODO: Fix when proper policies and access control are implemented
+    chown 1000 root /sys/class/nanohub/nanohub/mode
```
MagneFire, line 29: These look like they came from `catfish`. Can you remove them?

meta-triggerfish/recipes-core/base-files/base-files/fstab
```diff
@@ -0,0 +1,5 @@
+/dev/root            /                    auto       defaults              1  1
```
MagneFire, line 1: This looks like the default `fstab` (https://github.com/openembedded/openembedded-core/blob/kirkstone/meta/recipes-core/base-files/base-files/fstab) since you aren't mounting the `persist` partition. Can you remove this and check if everything is still ok.

MagneFire (commented):
Thank you for this work! It's great to see these new ports coming in!

I've made some minor deletion comments. Additionally, can you also remove the `fastboot.jpg` and `recovery.jpg` files?

MagneFire (approved):
Perfect! Thanks!

PR #139, opened 2022-12-19
Add support for Polar M600 (pike)

dodoradio:
This adds support for the Polar M600, an android wear sports watch that looks like an oversize fitbit. 
Like `mooneye`, this is based on an mtk2601. Unlike mooneye, this has working USB.
The port was initially based on `mooneye` but has mostly been moved over to a `narwhal` base. 

Currently working:
- Display
- Touch
- USB (in all modes)
- Heartrate and tilt-to-wake

Not working:
- Bluetooth
- GPS
- probably other things that I'm not aware of

I have not tested a full install on this watch as it's not supported on `mooneye` with the same SOC. This is something I will investigate later. 

less relevant notes:
- Various sources say that the display should be transflective, similar to `tetra` (sony smartwatch 3) or `bowfin` (moto 360 sport), however, I cannot confirm this, as I haven't yet seen the display fully turn off the backlight. 
- The power key is on the side; the button on the front is  customisable in android. Fastboot can be accessed by holding the front button while powering on and navigating the menu with the power key. The front button also has a white backlight accessible as `/sys/class/leds/button-backlight`. This has no brightness control, only turning off and on. 
- The UI takes a very long time to start on this device. This is definitely worth investigating


meta-pike/recipes-kernel/linux/linux-pike_o.bb
```diff
@@ -0,0 +1,33 @@
...
+S = "${WORKDIR}/git"
+B = "${S}"
+
+do_install:append() {
+    rm -rf ${D}/usr/src/usr/
+}
+
+BOOT_PARTITION = "/dev/mmcblk0p7"
+inherit mkboot old-kernel-gcc-hdrs
+inherit mkboot old-kernel-gcc-hdrs
```
MagneFire, line 33: There's a duplicate line here.

meta-pike/recipes-kernel/linux/linux-pike_o.bb
```diff
@@ -0,0 +1,33 @@
...
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://COPYING;md5=d7810fab7487fb0aad327b76f1be7cd7"
+COMPATIBLE_MACHINE = "pike"
+
+# Use an older version of gcc (gcc >= 9 doesn't boot.)
+DEPENDS:remove = "virtual/${TARGET_PREFIX}gcc"
+DEPENDS += "virtual/${TARGET_PREFIX}gcc8"
+
+SRC_URI = "git://android.googlesource.com/kernel/mediatek;branch=android-mediatek-pike-3.10-oreo-wear-dr;protocol=https \
+    file://0002-Fix-compilations-warnings.patch \
```
MagneFire, line 16: Can you fix the ordering of patches here?

meta-pike/recipes-android/android/android_pike-o.bb
```diff
@@ -0,0 +1,41 @@
+inherit gettext
+
+SUMMARY = "Downloads the Polar M600 /system and /usr/include/android folders and installs them for libhybris"
+LICENSE = "CLOSED"
+SRC_URI = "https://dl.dropboxusercontent.com/s/8b9t2renrxbl4gq/hybris-o-msm8909.tar.gz;name=hybris"
```
MagneFire, line 5: This is the same one used for other Oreo bases right? Maybe it would make sense to add a 'general' recipe for it at https://github.com/AsteroidOS/meta-asteroid/tree/master/recipes-android/android?

Obviously this could be done in another PR.

MagneFire (commented):
This PR generally looks good. Thanks for that!

One minor comment would be to remove the unused patches in `meta-pike/recipes-kernel/linux/linux-pike/`.

meta-pike/recipes-android/android/android_pike-o.bb
dodoradio: this is probably work that will go in meta-asteroid (similar to the msm8909 platforms for pie) so I'm not going to do anything here for now

MagneFire (approved):
Looks good now!
Thanks for your work on this platform!

PR #140, opened 2022-12-26
Add support for the Casio WSD-F20 (ayu/koi)

dodoradio:
This adds support for the `Casio Pro Trek Smart WSD-F20` (codenamed `ayu`). The layer is called `koi` because the kernel and most of the platform is shared with the `WSD-F10`, the `Casio Smart Outdoor Watch`. However, this has only been tested on a WSD-F20A, so I cannot actually say for sure that this is supported. 

This device is unique in a number of ways, including:
- Samsung Exynos 3250 dual-core processor
- layered reflective LCD [1]
- Charging light
- Size (this is the biggest WearOS watch)
- terrible hardware defects (described below)

# A series of disclaimers
This watch is both problematic and expensive. According to various forum threads, the following failures are embarrassingly common:
- Bricking of various sorts  (device gets stuck at Casio logo, or at spinning dots WearOS boot screen)
- touchscreen failures (these seem to be caused by delamination of some sort)

Casio does not provide extended warranty for these failures. The watches are also hilariously slow under WearOS, to the point that the second hand can be seen lagging on some watchfaces. Despite all this, used units still sell between $100-$200 in different markets. 

My personal example bricked itself while I was testing AsteroidOS, and no longer boots into android unless reflashed with a live dump. This also affects Asteroid: The environment variable `LD_BIND_NOW=1` needs to be set for all applications linked against Qt, or they exit with an illegal instruction error. This seems to be caused by the dynamic linker incorrectly mapping the qtbase library, but the issue is completely incomprehensible, and may be caused by a kernel or hardware bug. Just before the watch bricked itself, on one occasion I did get UI to work without the environment variable being set, suggesting that this may be related to the bricking. Because this issue is so unclear, I am not adding the environment variable to the build system until we can clarify on at least one other device.  I have currently manually added this variable to `/etc/systemd/system.conf`
.
# The port status
Currently, the following works, provided that I set `LD_BIND_NOW`:
- Display
- Touchscreen
- Vibration
- Bluetooth
- WLAN
- tilt-to-wake

The following does not work:
- GPS (only present on ayu, not present on koi) 
- Audio
- Compass (SensorFW seems to fail at some calibration stage, this needs to be checked)
- the second LCD [1]

USB is not externally accessible. On WSD-F20, this can be accessed by removing the four hex screws from the front, removing the whole front bezel and the `CASIO` logo plate. USB is available as two gold contacts, with D+ on the left and D- on the right. A custom cable is needed for this, and I have had no success with designing a dock for this. 

[1] This watch has a reflective segment LCD layered on top of the primary colour LCD. This LCD tells the time and contains a small dot-matrix which is used in WearOS to display the date and sensor data. Although it handles all the sensors on the device, the LCD portion of the [driver](https://android.googlesource.com/kernel/exynos/+/refs/heads/android-exynos-koi-3.18-oreo-dr-wear/drivers/misc/MultiSensors.c) is quite rudimentary, just copying over data that is written to it. @MagneFire and I have been attempting to reverse engineer an application that writes to this display, but we have only managed to set the time in minutes and hours, and to invert the colours of the display, with no success setting seconds, date or any of the additional functions.


dodoradio:
There seems to be an issue here, if usb_g_android is enabled, ssh works but adb does not. if it is disabled, adb works but ssh does not.. I'll have to investigate this further

Lisra-git:
Hey, just found this issue and can't tell how much this help!
I'm on my way since a year to fight against mistakes made by Casio making any reset to factory watch useless as they won't pass setup, I got new factory files to flash and was searching for the USB connection, i'll try this and report back, Having an alt OS on this is also probably the best for it to die properly, can't wait to see how this go!

Lisra-git:
and what's your setup to connect to USB, seem like it just won't detect? any special mode?

dodoradio:
Hi, the USB access requires disassembly. I plan to release proper instructions soon, but we're currently still figuring things out over on the matrix chat

dodoradio:
Please can you ask these questions on the matrix chat? The instructions here are intentionally brief, as the flashing process is unptedictable and I'd like to be able to provide one-on-one tech support until I write good instructions; this PR is not a good place for that tech support

dodoradio:
https://asteroidos.org/contact/

eLtMosen:
> I tried to search for it but I can't find it sweat_smile
We are at #asteroid-git:matrix.org

This is what i did to my Ayu.
Red is USB green and Black is USB white in my solder job here

IMG_0669

dodoradio:
With massive help from @beroset and @MagneFire I have managed to reverse engineer nearly all of the segment LCD functions, as far as I understand. Currently commands (which are all 7 bytes long) are known for:
- time setting
- display colour inversion (sets whether the segments are black on a white background or vice-versa)
- 12/24 hour timekeeping modes
- shutdown into 'timekeeping mode' (where the main system shuts down while the time and date is still shown on the segment LCD. this command only applies after the main display is turned on and off. this probably makes use of the backlight driver)

all of the above have been implemented in https://github.com/AsteroidOS/lcd-tools/pull/3

The secondary LCD is made up of:
- a dot martix capable of displaying 4 characters, graphs, and a limited selection of symbols
- a four digit numerical display with colon separator
- a `p` symbol to indicate am/pm in 12 hour format
- a 4 digit display with colon separator
- two lines above and below the central display, which are enabled in sensor display and timepiece mode
- a battery charge symbol, which turns on when watch is on charger. 

ic_launcher

normally, the day of week and day of month are displayed on the dot matrix, hours and minutes on large numerical display and seconds on lower display. 

The segment LCD also supports a series of special modes that display sensor data:
- compass - displays an alphabetical compass bearing in the dot matrix, numeric bearing in central display and time in lower display. also shows calibration symbol in dot matrix if compass is uncalibrated
- altimeter - displays altimeter history graph in dot matrix, current altitude in central display and time in lower display
- barometer - displays pressure history graph in dot matrix, current pressure in central display and time in lower display
- tide chart - can be set to display either `◠ high` or `◡ low` in dot matrix, time of tide event in main display and current time in lower display
- myself meter - displays `kcal` in top display, kilocalorie reading in center display and time in lower display

The sensor segment commands seem to hijack the backlight, which can render the watch inoperable until a hard reboot or the correct reset commands.

dodoradio:
some testing by @eLtMosen and further testing by me indicates `LD_BIND_NOW=1` is always needed in a full install, and is not needed in a temporary install

meta-koi/conf/machine/koi.conf
```diff
@@ -0,0 +1,20 @@
+
```
FlorentRevest, line 1: Unnecessary empty line up there

meta-koi/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
```diff
@@ -0,0 +1,27 @@
...
+mount --bind $BOOT_DIR/system/vendor/firmware/  /system/vendor/firmware/
+mount --bind $BOOT_DIR/vendor/firmware          /vendor/firmware
+mount --bind $BOOT_DIR/nvdata                   /nvdata
+mount --bind $BOOT_DIR/firmware                 /firmware
+
+dd if=/nvdata/MACWLAN of=/proc/wifi/mac
+
+ip link set up dev wlan0
+sleep 1
+ip link set down dev wlan0
```
FlorentRevest, line 27: Is this needed for the MACWLAN overwrite to be effective ?
If this is, could you add a comment describing why it's there ?

meta-koi/recipes-android/android/android_koi-o.bb
```diff
@@ -0,0 +1,47 @@
...
+    cp -r include/* ${D}${includedir}/android/
+
+    install -d ${D}${libdir}/pkgconfig
+    install -m 0644 ${D}${includedir}/android/android-headers.pc ${D}${libdir}/pkgconfig
+    rm ${D}${includedir}/android/android-headers.pc
+    cd ${D}
+    ln -s system/vendor vendor
+}
+
+# FIXME: QA Issue: Architecture did not match (40 to 164) on /work/dory-oe-linux-gnueabi/android/lollipop-r0/packages-split/android-system/system/vendor/firmware/adsp.b00 [arch]
```
FlorentRevest, line 40: I think you could fix this with one of the `INSANE_SKIP`, maybe if you add arch after already-stripped l15 ?

meta-koi/recipes-android/android-init/android-init_1.0.bbappend
```diff
@@ -0,0 +1,11 @@
+FILESEXTRAPATHS:prepend:koi := "${THISDIR}/${PN}:"
+
+SRC_URI:append:koi = " file://nonplat_property_contexts \
+    file://plat_property_contexts"
+
+do_install:append:koi() {
+    install -m 0644 ${WORKDIR}/nonplat* ${D}/
+    install -m 0644 ${WORKDIR}/plat* ${D}/
```
FlorentRevest, line 8: I'm curious why these files were needed ?
I see you boot your kernel cmdline contains selinux=0 already so selinux labels shouldn't be needed in principle ?

meta-koi/recipes-android/android/android/audio_policy.conf
```diff
@@ -0,0 +1,59 @@
+# Global configuration section: lists input and output devices always present on the device
```
MagneFire, line 1: This file is specific to `narwhal` can you remove this?

meta-koi/recipes-android/brcm-patchram-plus/brcm-patchram-plus/patchram.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
MagneFire, line 1: The `brcm-patchram-plus` service might not be needed on `koi` since we can use the bluebinder and Android Bluetooth service combination. Can you remove these?

meta-koi/recipes-core/firmwared/firmwared/firmwared.service
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
MagneFire, line 1: Is this service file used? I remember that this was specific to `sawfish`.

MagneFire (commented):
This PR looks good in general, there are just some minor changes I think you could apply here.

meta-koi/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
dodoradio: I'm not sure. I believe most of this is left over from whatever watch it's based on (ray?). I haven't looked into testing/cleaning these up yet because there were bigger issues, such as the perplexing `LD_BIND_NOW` requirement up until now.

meta-koi/recipes-core/psplash/psplash_git.bbappend
```diff
@@ -0,0 +1,6 @@
+SRC_URI:append:koi = " file://0002-Disable-double-buffering.patch"
```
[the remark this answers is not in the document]
dodoradio: koi does a show a boot logo. It does, however, seem to behave in a slightly different way to other watches.

dodoradio:
after some light debugging, I am sure that the compass issue is caused by a similar problem to the one seen with the Qt libraries. the issue can be worked around by using the temporary installation method, but it would be great if we figured out the actual cause and an actual solution.

meta-koi/recipes-android/android-init/android-init_1.0.bbappend
dodoradio: koi seems to boot correctly with no files here - should I just delete the files and the bbappend?

FlorentRevest: Okay fair enough

meta-koi/recipes-android/android/android_koi-o.bb
FlorentRevest: Alright! Maybe we could add a comment clarifying that then ?

meta-koi/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
FlorentRevest: Okay, it's not the end of the world either, it's just unfortunate to wait for 1 second in every boot if we don't even need to :)

dodoradio:
The entire LCD interface, at least everything used from wearos, has been reverse engineered. code is at https://github.com/dodoradio/lcd-tools/

MagneFire (commented):
First batch of comments, I'll have to look at it some more another day.
(Ran out of time to review this PR, sorry)

meta-koi/recipes-android/android/android_koi-o.bb
dodoradio: I'm not adding this here, unless this has already been added to all other recipes. I think that should be one commit that adds the clarification to all affected ports.

meta-koi/recipes-core/psplash/psplash_git.bbappend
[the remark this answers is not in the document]
dodoradio: the reason I wrote this initially is that the logo seemed to only show when I had the display driver set up correctly. This was probably just a coincidence.
Another quirk is that the system spends ages at the casio boot logo, and only shows the psplash asteroid logo for a split second before the launcher takes over. Perhaps this points to some inefficiencies in the early boot process, but the overall boot-to-ui time seems comparable to other watches, so I'm not making it a priority. 

back on topic: I'm removing this and marking this as resolved, since I've removed this file. behaviour is identical without it.

meta-koi/recipes-android/android/android_koi-o.bb
```diff
@@ -0,0 +1,46 @@
+inherit gettext
+
+SUMMARY = "Downloads the Snapdragon Wear 2100/3100 /usr/libexec/hal-droid and /usr/include/android folders and installs them for libhybris"
+LICENSE = "CLOSED"
+
+SRC_URI = "https://dl.dropboxusercontent.com/s/4d8pkmd62rxl464/hybris-o-msm8909.tar.gz;name=hybris"
```
MagneFire, line 6: Thanks for the reminder of 'tweak the tarball'. I've created an issue prompting for the other ports to be updated as well: https://github.com/AsteroidOS/meta-smartwatch/issues/172

MagneFire (approved):
PR looks good, thank you for finalizing it!

I would recommend to split a commit like https://github.com/AsteroidOS/meta-smartwatch/pull/140/commits/156a1de4469fda38e725f10ad70895dffab0084e though, as it contains changes that have nothing to do with Bluetooth, but are not needed for some other reason.

Also, please don't forget to add the `koi` machine to https://github.com/AsteroidOS/asteroid/blob/master/prepare-build.sh#L17 such that nightlies are built as well.

meta-koi/recipes-android/android/android_koi-o.bb
dodoradio: ah, makes sense. I'll see what I can do with the magic of interactive rebasing

MagneFire: But as mentioned, this is for future reference only. The PR is fine, but for future PRs I would recommend to split those kind of commits to a `android-init: Remove non-existent services` and `linux: Fix Bluetooth`

Issue #131, opened 2022-11-06
[beluga] No audio drivers


MagneFire:
The kernel drivers for the audio subsystem isn't packaged in the kernel. Instead they are available under a different repository: https://android.googlesource.com/kernel/msm-extra/+/refs/heads/android-msm-beluga-4.9-pie-wear-mr2.
The compiled kernel modules are located on the `vendor` partition, but these cannot be loaded since they mismatch due to kernel configuration changes. We should find a way to compile these kernel modules as well. Probably by means of a separate recipe.

wannaphong:
sound: https://android.googlesource.com/kernel/msm/+/refs/heads/android-msm-beluga-4.9-pie-wear-mr2/sound/
wifi: https://android.googlesource.com/kernel/msm-modules/wlan/+/refs/heads/android-msm-beluga-4.9-pie-wear-mr2

PR #141, opened 2022-12-26
beluga: Performance improvements, wlan, audio and Bluetooth fixes.

MagneFire:
This adds support for WiFi and audio on `beluga`.

The WiFi and audio kernel modules are not available by default in the kernel source, they are provided in other repositories. This means that they need to be compiled out-of-tree.
The WiFi kernel module is loaded by a systemd service as it doesn't appear to work early in the boot process.
The audio kernel modules are loaded via the `android-init` scripts, this is also the place where we boot the aDSP. After which the pulseaudio droid module will work.

It needed some changes to `make-mod-scripts`. For now I've made these changes only for `beluga` but we might want to move this to the `meta-asteroid` layer instead.

This PR also fixes the issue where the Bluetooth mac address would be reset to some random mac address when the watch is rebooted. This is achieved by starting the `btnvtool` binary.

This imports some performance improvements from a previous `catfish` PR: https://github.com/AsteroidOS/meta-smartwatch/pull/109

This fixes: https://github.com/AsteroidOS/meta-smartwatch/issues/131


FlorentRevest (approved):

FlorentRevest:
Excellent, I should really get a beluga! It's the first time I see this kind of out of tree modules but that's cool :) the more out of tree modules vendor kernels will have the easier it will be to update the base kernel

PR #149, opened 2023-01-21
ray/firefish: Bump base to Pie.

MagneFire:
This is basically a carbon copy of the work done to get `catfish` working on a pie base: https://github.com/AsteroidOS/meta-smartwatch/pull/109

This fixes NFC and GPS on at least `ray`.

It does introduce some strange horizontal lines and there appear to be some ghost touches. This is the reason why this PR will be made draft for now.


FlorentRevest (approved):

MagneFire:
The mentioned issues should be solved with this change: https://github.com/AsteroidOS/meta-asteroid/pull/178

I've noticed another issue though: The screen wakes immediately after sleeping, an interrupt is waking the display. I'll look into fixing this before getting this PR merged.

FlorentRevest:
I assume this is now totally stale - please close if appropriate

MagneFire:
Indeed, super stale, may revisit later, but it's probably easier to use this as inspiration instead of fixing the current state of this PR.

PR #163, opened 2023-02-25
dory: Improve Bluetooth and MCE.

MagneFire:
Switch to the `bluetooth-rfkill-event` and adjust the `mce` Always-on-Display configuration.

This also defines `tosleep` which makes flashing the patchram always succeed, instead of sometimes hanging.
It's worth noting that Android does something similar here: https://android.googlesource.com/platform/hardware/broadcom/libbt/+/f219785e8a4e011abe5488d22f3d72b6698ce45a/src/hardware.c#689

This reduces the brightness of Always-on-Display to the minimum when the normal brightness is set to less than 50% otherwise the normal behavior is kept (Keep the brightness at 28%).


FlorentRevest:
> This also defines `tosleep` which makes flashing the patchram always succeed, instead of sometimes hanging. It's worth noting that Android does something similar here: https://android.googlesource.com/platform/hardware/broadcom/libbt/+/f219785e8a4e011abe5488d22f3d72b6698ce45a/src/hardware.c#689

:star_struck: 

Great job! This one has been seriously annoying for a long long long time.

FlorentRevest (approved):

PR #164, opened 2023-02-25
minnow: Cleanup port.

MagneFire:
This PR cleans up the `minnow` port by removing porting remnants from other ports (mostly `sturgeon`).


FlorentRevest:
The terrible truth about check-config is that 95% of it is completely irrelevant, and the options which we actually really care about are probably not covered...

Anyway. Cool cleanup, thank you! :)

FlorentRevest (approved):

PR #165, opened 2023-02-26
add support for `medaka` Casio Pro Trek Smart WSD-F21hr

dodoradio:
This is the successor to `ayu`, the Casio WSD-F20. Compared to that, this watch adds a heartrate sensor, with a different segment LCD and slight differences in styling - overall, the two look very similar and even use the same primary display panel, but they are internally based on completely different platforms, with this one running the 64 bit Exynos 7270. 

obsolete commentary from three years ago:
> This port is not ready to merge for 2 reasons: 
> Firstly, the commit history is a total mess, since I removed and re-added a lot of beluga base in various commits. I'm not sure if it's worth slightly cleaning up but keeping all these changes, so that they act as a good log of the porting process, or if it's worth completely reworking the git history into something that would seem to be a more logical progression. 
> Secondly, the port currently uses a gross multilib hack to build a 64 bit kernel with a 32 bit userspace, which is necessary because all the android libraries are 32 bit. The multilib hack (from my minimal research) is the most correct solution for our build system, but it means that `lib32-` must be prepended to any package name that is build - ie. `lib32-asteroid-image` instead of `asteroid-image` and all packages come out as `lib32-[packagename]` as well, such as `lib32-lcd-tools_+git.....ipk`. This isn't ideal, but I'm not sure how best to fix it. 

## What works: 
- Display
- Touch
- Sensors (heartrate, tilt-to-wake and compass have been tested)
- Vibrator

## What doesn't work:
- GPS
- too expensive
- aforementioned 32 bit build issue
- USB in any mode other than adb

## Untested:
- microphone

Flashing requires identical disassembly steps to `ayu`.

it also seems like this port could be mostly shared for the WSD-F30, whenever one of us is lucky enough to get one of those rare and expensive watches in hand


dodoradio:
This has been redone, with a proper 32bit userspace and kernel recipe arch overridden using a technique made up by @argosphil 
I was inspired to take another crack at it by kido's recent work on aurora, but since that watch uses clang, very little of the actual 32-64bit solution was transferable
These are the kernel hacks:
- The architecture for the machine is set to arm(32) and the
  kernel has its arch overridden to aarch64
 - This means the kernel recipe needs 64bit gcc and binutils. These
   are not normally available like this, so they get brought in by the
   weird recipes in recipes-devtools.
- 32bit busybox mount doesn't like something about this 64 bit kernel,
  and fails to mount loop devices. This is hacked around by just
  including util-linux-mount in the ramdisk, so asteroidos.ext4 can
  mount (this watch doesn't seem to support having the asteroidos rootfs
  flashed directly to userdata, so the loop mount seems necessary)

The rest of this port is fairly conventional.

FlorentRevest (approved):
Nit: the initramfs adb logic still refers to the OPPO Watch.

FWIW, I also dislike how all our initramfs scripts are all different from one another now but it's not something I will block your PR on since it's existing tech debt. But everytime I see yet another custom initramfs it pains me a little bit :p

dodoradio:
nit fixed. 
yeah, the ramdisks are a bit of a mess. On this watch, it's just copied from beluga. It seems they do at least two separate things:
- mount the boot partition, bind ld.config.28.txt, bind mount the system partition. 
    - (worryingly, these partition mounts are hardcoded in the script)
    - hoki doesn't do this
- start adbd if asteroidos.ext4 isn't found
it'd be great if this functionality was split out and the partitions weren't hard-coded.  I might look into that later.

FlorentRevest (approved):

MagneFire (approved):

MagneFire:
Thanks for picking this back up!

Great work!

MagneFire:
Guessing this will need to be added to https://github.com/AsteroidOS/asteroid/blob/2.0/prepare-build.sh#L17 as well?

Issue #172, opened 2023-03-29
Android O based ports: Update the system tarball


MagneFire:
Currently, the `koi` port contains a newer version of the Android O tarball. This contains fixes with regards to the TLS region.
It's essential that the existing Android O based ports are updated as otherwise they will crash on newer versions of libhybris due to corruption of the TLS region.

PR #178, opened 2023-04-02
Use generic Android O patched libaries and fix Android Pie inconsistency.

MagneFire:
Switches all Oreo based ports to the generic tarball in the `meta-asteroid` repo.

Additionally, it fixes naming consistency issues with the Pie ports.

This depends on https://github.com/AsteroidOS/meta-asteroid/pull/144.

This fixes https://github.com/AsteroidOS/meta-smartwatch/issues/172.


dodoradio:
could you please mention in the `koi` commit that it does not actually have the msm8909 soc, and potentially mention this in a comment in the relevant files? it's not necessarily a 'hack' but it is worth clarifying why the port makes reference to a platform that isn't actually used in the watch. I realise I didn't previously clarify this, but it's probably worth doing while you're making these changes

MagneFire:
Now tested on: `catfish`, `beluga`, `narwhal`, `pike` and `ray`.

@dodoradio I've changed the naming of the recipes to just: `armv7-oreo`, `armv7-pie32` and `armv7-pie64`. The description summary for the oreo recipe has been updated to be generic but with some extensions for `Wear 2100/3100`.

dodoradio:
name changes look good, thank you for cleaning those up

PR #183, opened 2023-04-29
Add support for hoki (Fossil Gen 6)

MagneFire:
This adds support for the Fossil Gen 6 (Technically it could work on the entire Fossil Gen 6 platform, but this has not been tested).

The following works:
- Display
- Touchscreen
- Vibrator
- USB: Only ADB (usb-moded is broken)
- Bluetooth
- Buttons & Crown

The following is broken:
- Always-on-Display: Entering leaving AoD works fine. However the screen doesn't update.
- Sensors: The modem doesn't boot. It leads to a watchdog crash. (No heartrate, tilt-to-wake, ambient light sensor)
- NFC: Requires a functional modem
- GPS: Requires a functional modem
- Audio: Requires a kernel driver (https://github.com/fossil-engineering/kernel-msm-fossil-extra-cw-audiokernel) and likely a functioning modem as well.

I've documented some of my findings for this port over at: https://gist.github.com/MagneFire/fe4e91ca658ef6d3ca79ce4e344bd914#hoki

This port is currently using an Android 9 base (from an XDA dump). I intend to update that to an Android 11 base in an attempt to get the modem to work.

This PR depends on https://github.com/AsteroidOS/meta-asteroid/pull/153.


meta-hoki/recipes-nemomobile/ngfd/ngfd/51-ffmemless.ini
```diff
@@ -0,0 +1,208 @@
+[ffmemless]
```
FlorentRevest, line 1: I wonder if this file should be moved to meta-asteroid at some point. But not blocking this PR on that.

PR #193, opened 2023-05-19
narwhal: Add hands-timesync

dodoradio:
despite this existing as an asteroid repo for over a year now, it's not actually been added as a recipe. I think this is a sign of anti-narwhal prejudice. 

This adds the recipe, and adds it to the narwhal image. 
it also removes the tz_minutes reset from init.rc: This was previously needed to make sure that the time would not change between reboots when it was being read back from the hands controller on bootup. This is no longer relevant, as asteroid now supports timezones (so the 0 reset would actually be harmful) and hands-timesync takes care of setting this value correctly.

It is worth checking how I've got the systemd units set up here. Unlike catfish and ayu, where the LCDs only display the time, the time set to the hands will be written to the main system after a reboot. For this reason, the time setting (as is convention on Linux) is restricted to root by making the systemd service and timer run as root, and keeping the permissions for the files /sys/devices/sop716/time and /sys/devices/sop716/tz_minutes locked down.

### DIFF
--- a/meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
+++ b/meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
@@ -1,0 +1,20 @@
+DESCRIPTION = "Utility to set correct time on LG Watch W7"
+HOMEPAGE = "https://github.com/AsteroidOS/narwhal-hands-timesync.git"
+LICENSE = "GPL-3.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=84dcc94da3adb52b53ae4fa38fe49e5d"
+
+SRC_URI = "git://github.com/asteroidos/narwhal-hands-timesync.git;branch=main;protocol=https"
+SRCREV = "${AUTOREV}"
+PR = "r0"
+PV = "+git${SRCPV}"
+S = "${WORKDIR}/git"
+DEPENDS = "qtbase"
+inherit cmake_qt5
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install:append() {
+    install -d ${D}/etc/systemd/system/timers.target.wants/
+    ln -s /usr/lib/systemd/system/hands-timesync.timer ${D}/etc/systemd/system/timers.target.wants/hands-timesync.timer
+}
+
+FILES:${PN} += "/usr/lib/systemd/system/ /etc/systemd/"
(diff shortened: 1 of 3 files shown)


meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
```diff
@@ -0,0 +1,22 @@
...
+inherit cmake_qt5 pkgconfig
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install:append() {
+    install -d ${D}/usr/lib/systemd/system/
+    install -d ${D}/etc/systemd/system/timers.target.wants/
+    ln -s /usr/lib/systemd/system/hands-timesync.timer ${D}/etc/systemd/system/timers.target.wants/hands-timesync.timer
+}
+
+FILES:${PN} += "/usr/lib/systemd/system/ \
```
MagneFire, line 21: Unnecessary line break here, can you make this a one liner?

meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
```diff
@@ -0,0 +1,22 @@
...
+SRCREV = "${AUTOREV}"
+PR = "r0"
+PV = "+git${SRCPV}"
+S = "${WORKDIR}/git"
+DEPENDS = "qtbase"
+inherit cmake_qt5 pkgconfig
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install:append() {
+    install -d ${D}/usr/lib/systemd/system/
```
MagneFire, line 16: Isn't this path created by CMake?

meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
```diff
@@ -0,0 +1,22 @@
...
+LICENSE = "GPL-3.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=84dcc94da3adb52b53ae4fa38fe49e5d"
+
+SRC_URI = "git://github.com/asteroidos/narwhal-hands-timesync.git;branch=main;protocol=https"
+SRCREV = "${AUTOREV}"
+PR = "r0"
+PV = "+git${SRCPV}"
+S = "${WORKDIR}/git"
+DEPENDS = "qtbase"
+inherit cmake_qt5 pkgconfig
```
MagneFire, line 12: `pkgconfig` isn't required here since it only depends on Qt.

MagneFire (commented):
Minor comments, everything else looks good!

meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
dodoradio: it should be. I was working on the basis of lcd-tools, and I stuck with the format used there. will remove.

dodoradio:
resolved all comments

PR #200, opened 2023-07-16
libhybris: Use same version for all platforms.

MagneFire:
This reduces the burden of having to maintain various versions of libhybris for different platforms. The only exception is the use of reporting GLES 2.0 instead of 3.0 for the platforms that need it.


FlorentRevest:
Very cool, thanks a lot :) We shouldn't have left that much tech debt accumulate. But kudos for cleaning it!

PR #203, opened 2023-07-31
koi: set permissions for LCD in init.rc

dodoradio:
This is needed for ceres to be able to set time and change settings on the LCD.


MagneFire:
Thank you :)

PR #205, opened 2023-08-02
dory: brcm-patchram-plus: Remove patchram service.

MagneFire:
Since `bluetooth-rfkill-event` the patchram service is no longer used on `dory`.

Depends on: https://github.com/AsteroidOS/meta-asteroid/pull/172


beroset (approved):
This looks good and has been tested successfully on `dory`.

PR #206, opened 2023-08-02
wren: Enable WLAN utilities and autoload kernel module.

MagneFire:
Implements the changes needed to get WiFi to work.


eLtMosen (commented):
Compiling with those changes just now to test, thanks for the swift action!

PR #209, opened 2023-08-02
sawfish: linux-sawfish: Remove USB gadget function list.

MagneFire:
Instead use the default list provided by the android gadget driver.

Essentially it adds rndis mode to the supported functions list, fixing SSH mode.


eLtMosen:
Tested on sawfish just now! Nicey Nicey

FlorentRevest:
Great job :)

PR #219, opened 2023-08-25
linux-audio-modules-beluga: fix mickledore build

argosphil:
Avoid using a problematic GNU Make construct to allow building on mickledore.

I don't fully understand how `make` has changed and how it causes the problem, but there is code in `make` now that is reached in this specific situation.

Note that `asoc/Kbuild` includes `DEFINES += -DBUILD_TIMESTAMP=\"$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')\"`, which doesn't have any effect since it should be `CDEFINES`. Not modified to keep changes to a minimum.

### DIFF
--- a/meta-beluga/recipes-kernel/modules/linux-audio-modules-beluga/0002-Avoid-shell-expansion-in-recursively-expanded-variab.patch
+++ b/meta-beluga/recipes-kernel/modules/linux-audio-modules-beluga/0002-Avoid-shell-expansion-in-recursively-expanded-variab.patch
@@ -1,0 +1,25 @@
+From 7bcf29c0177330f64af7eeca32fefb0502ceeefe Mon Sep 17 00:00:00 2001
+From: Philip Russell <argosphil@murena.io>
+Date: Fri, 25 Aug 2023 12:26:10 +0000
+Subject: [PATCH] Avoid shell expansion in recursively-expanded variable
+
+GNU Make changed the behavior of that particular (avoidable and problematic)
+construct. Eagerly expanding the timestamp fixes the build on Mickledore.
+---
+ ipc/Kbuild | 3 ++-
+ 1 file changed, 2 insertions(+), 1 deletion(-)
+
+diff --git a/ipc/Kbuild b/ipc/Kbuild
+index b4a0f985..721526bd 100755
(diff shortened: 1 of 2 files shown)


FlorentRevest:
I'm happy to merge this. I don't think you need to waste energy root causing make changes :) it's a clean change and if it fixes your build I think we should just merge it.

FWIW, I had a quick look at https://android.googlesource.com/kernel/msm-extra/+/refs/heads/android-msm-beluga-4.9-pie-wear-mr2/ipc/ and it doesn't look like any c file refers to that macro so we could probably even remove it but I think your change is fine as is.

argosphil:
Oops, thought I'd responded here. We found what I believe is the root cause: the makefiles contain an unlimited blanket `export` statement, and that exports recursively-defined variables, which of course overrides their values in child makes and breaks everything. I'm not quite sure why it worked with the old `make`...

Anyway, it seems we still have sound problems on `mickledore`/`beluga`, so I've been meaning to investigate those. At some point I thought it was caused by the modules, but now I'm not so sure; it might be an intermittent problem.

MagneFire:
> Oops, thought I'd responded here. We found what I believe is the root cause: the makefiles contain an unlimited blanket `export` statement, and that exports recursively-defined variables, which of course overrides their values in child makes and breaks everything. I'm not quite sure why it worked with the old `make`...
> 

I've processed your findings here as well: https://github.com/MagneFire/meta-smartwatch-1/commits/update-to-mickledore. Depending on what you want we can apply the changes here or in https://github.com/AsteroidOS/meta-smartwatch/pull/186.

MagneFire:
> Anyway, it seems we still have sound problems on `mickledore`/`beluga`, so I've been meaning to investigate those. At some point I thought it was caused by the modules, but now I'm not so sure; it might be an intermittent problem.

For what it's worth, I've tested your changes on `kirkstone` and it all works fine (`quake` and `gpsp` work correctly).

argosphil:
> For what it's worth, I've tested your changes on `kirkstone` and it all works fine (`quake` and `gpsp` work correctly).

Thank you!

Let's apply them to the `kirkstone` branch, then! That should make transition easier.

argosphil:
superseded by #226

PR #227, opened 2023-09-15
[DRAFT] Initial `rubyfish` port

argosphil:
(Sorry, I've been dragging my feet on this for a while)

This is the initial port to `rubyfish`. It still needs some rebasing, of the kernel parts in particular, but I'd like to request an initial review and any hints or advice as to how to improve the PR so it can be accepted.

Status of the port:
* kernel boots
* adbd over USB works
* psplash works
* asteroid-launcher works
  * performance seems acceptable
* touchscreen works
* battery indication works
* sensors
  * ALS seems miscalibrated (screen is too dark to see in dark rooms)
  * heart rate works
  * SpO2 works using `/vendor/bin/nanotool`, no AsteroidOS integration yet?
  * barometer works using `/vendor/bin/nanotool`, no AsteroidOS integration yet?
  * others are untested
* NFC is untested
* GPS is untested
* Bluetooth works
* WiFi works, but appears to scan with a random MAC address which means the connection will not be reused across reboots
* Audio does not work
  * it is possible to get audio output but this requires manual (command-line) initialization of the adsp and aw88230 i2c-controlled i2s amp
  * mic hasn't been tested yet. 
* secondary LCD display requires manual (command-line) initialization
* tilt-to-wake untested
* tap-to-wake untested
* AoD (non-LCD) doesn't update (same issue as `hoki`?)
* vibrator used to work, but no longer does...

TODO:
* renumber/rework Linux patches
* remove obsolete device trees etc. from Linux directory
* test on more than one watch
* port to `rover`
* fix AoD


FlorentRevest:
> This is the initial port to `rubyfish`. It still needs some rebasing, of the kernel parts in particular, but I'd like to request an initial review and any hints or advice as to how to improve the PR so it can be accepted.

Overall it's a pretty clean port so kudos :)

A couple of things that could be improved are:
- in your kernel and drivers patches, make sure you don't just explain what the patch does but also why. usually the patch is quite straightforward and the reader can understand what the patch does. But it's not at all clear why you'd do things like disabling interrupts for example. And even if the reason is "dirty hack to fix compilation" or something like this, this is still tremendously useful context for whoever will look into this next :)
- It's best to avoid binary blobs in git repos because they stay forever and make repos unnecessarily heavy. I don't see the need for meta-rubyfish/images/rubyfish.png, is this something you intend to send to the asteroidos.org repo in the end ?
- I don't want to punt this on you because it's already a mess but I dislike that we have this init.machine.sh copied all around. I think the main init.sh script should evolve to support configfs or functionfs cleanly and init.machine.sh should stay as minimal as possible
- maybe it would be nice to fold the two vibrator commits. But you also mentioned that vibrator stopped working so maybe that's why you didn't fold them into one

> TODO:
> * renumber/rework Linux patches
> * remove obsolete device trees etc. from Linux directory

Yeah that was one of my questions. Doesn't the dts in the repo work ?

> * test on more than one watch
> * port to `rover`
> * fix AoD

Fwiw, stuff like aod shouldn't be a blocker for merging the initial support ;)

Charlyo:
Any news on the topic? Would love to have this running on my watch :)

argosphil:
Waiting for some pogo pins...

Charlyo:
Have you found a source for them? If you need an extra one or a hand, let me know :)

Issue #185, opened 2023-04-29
Modem crashes


MagneFire:
The modem isn't functional on `hoki` (used for sensors, GPS, NFC, Audio).

It's crashing as the watchdog detects a hang:
```
Apr 16 00:45:25 hoki kernel: Fatal error on the modem.
Apr 16 00:45:25 hoki kernel: modem subsystem failure reason: dog.c:1522:Watchdog detects stalled initialization.
```

Getting the log under `/sys/kernel/debug/ipc_logging/soc:qcom,smd:modem/log_cont` shows:
```
[     0.141965937/         0x9d54e85] qcom_smd_register_edge: soc:qcom,smd:modem
[    19.639723236/        0x20258a4b] qcom_smd_channel_reset: ch IPCRTR ed mpss
[    19.639798028/        0x20258fe6] qcom_smd_create_device: registering ch IPCRTR
[    19.640206778/        0x2025ae8f] qcom_smd_channel_set_state: ch IPCRTR ed mpss
[    19.640237299/        0x2025b0d8] qcom_smd_edge_intr: to APPS ch IPCRTR ed mpss
[    19.640246830/        0x2025b18f] qcom_smd_channel_set_state: ch IPCRTR ed mpss
[    19.640278393/        0x2025b3ee] qcom_smd_channel_get_rx_avail: h: 0x48 t: 0x0 ch IPCRTR ed mpss
[    19.640283340/        0x2025b44c] qcom_smd_channel_intr: pkt_size: 52 ch IPCRTR ed mpss
[    19.640285997/        0x2025b47f] qcom_smd_channel_get_rx_avail: h: 0x48 t: 0x14 ch IPCRTR ed mpss
[    19.640290320/        0x2025b4d2] qcom_smd_channel_recv_single: ret -11 len 52 ch IPCRTR
[    19.640293861/        0x2025b516] qcom_smd_channel_intr: fail ret -11 ch IPCRTR ed mpss
[    19.640295424/        0x2025b534] qcom_smd_edge_intr: to APPS ch IPCRTR ed mpss
[    19.640447611/        0x2025c09f] qcom_smd_channel_get_rx_avail: h: 0x48 t: 0x14 ch IPCRTR ed mpss
[    19.640498028/        0x2025c466] qcom_smd_channel_recv_single: len 52 ch IPCRTR ed mpss
[...]

PR #237, opened 2024-01-10
hoki: Fix sensors

MagneFire:
Provides the audio kernel modules to allow booting the modem.
Then start the required Android services to get sensors to work.

Audio isn't functional at this point, I suspect that this is because the Android R firmware is loaded on an Android Pie base, which may require a different audio routing.

This depends on https://github.com/AsteroidOS/meta-asteroid/pull/182.

This solves #185.


FlorentRevest:
Excellent, thank you MagneFire! :D

PR #239, opened 2024-02-13
emulator: Move qemux86-specific changes into separate machine meta layer

casept:
Previously, qemux86-related changes were strewn all over meta-asteroid, which went against the way all other machines are handled.

This change creates a new "smartwatch" machine, called emulator. It inherits most machine configuration from openembedded's qemux86 machine.

The patches, configs etc. are basically the same as they were in meta-asteroid, the interesting change is the creation of a dedicated machine.conf.


casept:
Oops, looks like I managed to break mouse support. Converting back to draft until that's fixed

FlorentRevest:
I see this is still marked as draft, presumably because you haven't gotten a chance to fix mouse support, is that right ?

This is the reason why I have not commented on https://github.com/AsteroidOS/asteroid/pull/281 and https://github.com/AsteroidOS/meta-asteroid/pull/188 and this one yet. 

FWIW, the changes look good to me overall. Are you sure that mouse support even worked before your refactoring ? If not, that's a strict improvement and we can move forward with it anyway. If something did break though, it would be good to figure it out before merging :)

casept:
Yes, I'm sure that the mouse support does work on the current master and is broken by my PR.

I think I've managed to narrow it down to a strange configuration installation issue - for whatever reason, when I try to override the config files in a .bbappend in `meta-emulator`, bitbake silently ignores the changes. If I make the prepend unconditional it works, but obviously breaks other machines.

casept:
> I think the idea of this is good. One thing that I wonder about is whether, by renaming from `qemux86` to `emulator` if we might lose some of the Yocto functionality that uses qemu. See https://docs.yoctoproject.org/dev-manual/qemu.html

Looks like this is not a problem as the MACHINE can be passed to `runqemu` explictly. E.g.:

```
$ runqemu emulator
runqemu - INFO - Running MACHINE=emulator bitbake -e  ...
runqemu - INFO - Continuing with the following parameters:
KERNEL: [/build/asteroid/build/tmp-glibc/deploy/images/emulator/bzImage]
MACHINE: [emulator]
FSTYPE: [ext4]
ROOTFS: [/build/asteroid/build/tmp-glibc/deploy/images/emulator/asteroid-image-emulator.rootfs-20250928160343.ext4]
CONFFILE: [/build/asteroid/build/tmp-glibc/deploy/images/emulator/asteroid-image-emulator.rootfs.qemuboot.conf]

runqemu - INFO - Setting up tap interface under sudo
runqemu - INFO - Network configuration: ip=192.168.7.2::192.168.7.1:255.255.255.0::eth0:off:8.8.8.8 net.ifnames=0
runqemu - INFO - Running /build/asteroid/build/tmp-glibc/work/x86_64-linux/qemu-helper-native/1.0/recipe-sysroot-native/usr/bin/qemu-system-i386 -device virtio-net-pci,netdev=net0,mac=52:54:00:12:34:02 -netdev tap,id=net0,ifname=tap0,script=no,downscript=no -object rng-random,filename=/dev/urandom,id=rng0 -device virtio-rng-pci,rng=rng0 -drive file=/build/asteroid/build/tmp-glibc/deploy/images/emulator/asteroid-image-emulator.rootfs-20250928160343.ext4,if=virtio,format=raw -usb -device usb-tablet -usb -device usb-kbd   -cpu IvyBridge -machine q35,i8042=off -smp 4 -m 256 -serial mon:vc -serial null -display none  -kernel /build/asteroid/build/tmp-glibc/deploy/images/emulator/bzImage -append 'root=/dev/vda rw  mem=256M ip=192.168.7.2::192.168.7.1:255.255.255.0::eth0:off:8.8.8.8 net.ifnames=0 oprofile.timer=1 tsc=reliable no_timer_check rcupdate.rcu_expedited=1 swiotlb=0 '

runqemu - INFO - Host uptime: 3729474.54
```

PR #246, opened 2024-06-22
hoki: linux-hoki: Disable system data verification

MagneFire:
This is not used with the current kernel and may cause kernel build issues as it requires libcrypto. libcrypto currently isn't amongst the build dependencies, adding doesn't appear to solve these issues.


FlorentRevest:
Those would be good to add to https://github.com/AsteroidOS/meta-asteroid/blob/master/scripts/check-config#L214 :)

PR #250, opened 2024-08-06
Add support for Yocto scartgap

MagneFire:


meta-triggerfish/recipes-kernel/linux/linux-triggerfish_p.bb
beroset, line 1: This change looks correct, but shouldn't we also change this line in `conf/machine/triggerfish.conf`?
```
PREFERRED_VERSION_linux = "3.18+pie"
```

meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
beroset, line 1: Probably a good fix, but the commit message doesn't tell why the change was needed and I can't find documentation on yoctoproject.org website for this ${systemd_system_unitdir} variable.

meta-sparrow/recipes-kernel/linux/linux-sparrow/0005-Create-copy-of-devfreq_trace.h.patch
```diff
@@ -1,4 +1,4 @@
-From 7e53114b1f8b917ef9f4fe1e1cf6908526b7ffc7 Mon Sep 17 00:00:00 2001
```
beroset, line 1: It's not clear why these hash values changed for a number of otherwise untouched patch files and only for sparrow.  Was this some devtool artifact?

meta-beluga/recipes-kernel/modules/linux-wlan-modules-beluga_p.bb
```diff
@@ -23,6 +23,6 @@ SYSTEMD_PACKAGES = "${PN}"
 SYSTEMD_SERVICE:${PN} = "wlan-module-load.service"
 
 do_install:append() {
-    install -d ${D}/lib/systemd/system/
-    cp ${WORKDIR}/wlan-module-load.service ${D}/lib/systemd/system/
+    install -d ${D}${systemd_system_unitdir}
```
beroset, line 26: Probably a good fix using `install` instead of `cp`, but I can't find documentation on ${systemd_system_unitdir} on yoctoproject.org.

beroset (changes requested):
I saw 4 items I wanted to ask about:

1. triggerfish went from Linux 3.18 to 4.9 (but the layer.conf still says 3.18)
2. narwhal now uses ${systemd_system_unitdir}
3. beluga now uses ${noarch_base_libdir}
4. sparrow patches seem to have a lot of git hash changes in patches

Questions are on the associated code bits.

meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb
MagneFire: Good point, I realise that this commit message is misleading, it doesn't actually change anything but use a provided variable instead of a hardcoded path.

I'll reword the relevant commit.

As for where this comes from, I generally look https://github.com/openembedded/openembedded-core/blob/scarthgap/meta/conf/bitbake.conf for useful variables that can replace hardcoded ones. systemd_system_unitdir is defined in that config file.

meta-sparrow/recipes-kernel/linux/linux-sparrow/0005-Create-copy-of-devfreq_trace.h.patch
MagneFire: Correct, this is just an artifact from a `devtool finish` iirc.

I can just remove those changes if you want me too?

meta-beluga/recipes-kernel/modules/linux-wlan-modules-beluga_p.bb
MagneFire: See previous comment ;)

MagneFire:
> I saw 4 items I wanted to ask about:
> 
> 1. triggerfish went from Linux 3.18 to 4.9 (but the layer.conf still says 3.18)
> 2. narwhal now uses ${systemd_system_unitdir}
> 3. beluga now uses ${noarch_base_libdir}
> 4. sparrow patches seem to have a lot of git hash changes in patches
> 
> Questions are on the associated code bits.

Thanks for reviewing this PR! I'll process the feedback tomorrow!

See answers to the relevant bits.

meta-sparrow/recipes-kernel/linux/linux-sparrow/0005-Create-copy-of-devfreq_trace.h.patch
beroset: Not necessary -- just wanted to make sure.

beroset (approved):

PR #254, opened 2024-12-23
Provide a means to easily reflash AsteriodOS

beroset:
This shows the AsteroidOS logo and checks to see if AsteroidOS is already present.  If it isn't we start the ADB daemon and tell the user to flash the ext4 image.  If it is present, the user has five seconds to press the middle button.  If the user presses the middle button during those 5 seconds, we start the ADB daemon; othewise it just boots normally.

### DIFF
--- a/meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
+++ b/meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
@@ -27,7 +27,7 @@ startadbd() {
 # u32 - tv_usec
 # u16 - type
 # u16 - code
-# s32 - value 
+# s32 - value
 # On this watch the middle button has a type of 0x74
 # and the bottom button is 0x10a
 # The top button is encoded in a different event.
@@ -42,9 +42,9 @@ waitmidbutton() {
 if [ ! -e $BOOT_DIR/etc/asteroid/machine.conf ] ; then
     startadbd
 else
-# wait for either 5 seconds or for the user to hit the middle button
+# wait for either 1 second or for the user to hit the middle button
 # If the timout expires, the watch boots normally; otherwise it starts ADB daemon
     trap "startadbd" SIGUSR1
     waitmidbutton $$ &
-    sleep 5
+    sleep 1
 fi
--- a/meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
+++ b/meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
@@ -21,6 +21,16 @@ startadbd() {
 
 # wait for the middle button to be pressed, and if it is,
 # kill the calling process group with SIGUSR1
+# The format of these events is documented here:
+# https://github.com/torvalds/linux/blob/master/include/uapi/linux/input.h#L28-L47
+# u32 - tv_sec
+# u32 - tv_usec
+# u16 - type
+# u16 - code
+# s32 - value 
+# On this watch the middle button has a type of 0x74
+# and the bottom button is 0x10a
+# The top button is encoded in a different event.
 waitmidbutton() {
     fmtstring='1/4 "%d." 1/4 "%d\t" 1/2 "%2x\t" 1/2 "%2x\t" 1/4 "%u\n"'
     result=$(hexdump -n16 -e "${fmtstring}" /dev/input/event0 | cut -f3 -)
--- a/meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
+++ b/meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
@@ -2,9 +2,9 @@
 
 BOOT_DIR=$1
 
-# Check if the AsteroidOS specific machine configuration file exists.
-# If it doesn't then we know that the userdata partition is mounted but doesn't contain a valid AsteroidOS root.
-if [ ! -e $BOOT_DIR/etc/asteroid/machine.conf ] ; then
+# Start the ADB daemon
+startadbd() {
+    trap - SIGUSR1
     echo MSG "Push asteroidos.ext4 via ADB" > /run/psplash_fifo
 
     # Setup for use with ConfigFS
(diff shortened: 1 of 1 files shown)


MagneFire (approved):

MagneFire:
Tested works fine! Thanks for providing this!

meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
```diff
@@ -17,4 +17,24 @@ if [ ! -e $BOOT_DIR/etc/asteroid/machine.conf ] ; then
     echo Fossil Gen 6 > /sys/kernel/config/usb_gadget/g1/strings/0x409/product
 
     /usr/bin/adbd
+}
+
+# wait for the middle button to be pressed, and if it is,
+# kill the calling process group with SIGUSR1
+waitmidbutton() {
+    fmtstring='1/4 "%d." 1/4 "%d\t" 1/2 "%2x\t" 1/2 "%2x\t" 1/4 "%u\n"'
```
MagneFire, line 25: This looks a bit cryptic, perhaps add a description of how the bytes are parsed?

Leaving it up to you. Feel free to merge regardless.

dodoradio:
I'm alright withj this being a temporary solution for hoki, but I would argue against merging this on other watches where only temporary install is available.
I'd suggest the boot image checks which partition it booted from (boot or recovery) and starts the adb daemon depending on that. Or we just provide a separate recovery image which always boots to adb. That's how SailfishOS does it.
the issue for me is that bootup on koi (casio wsd-f20, which also only uses 'temporary' install) is already incredibly slow and slowing it down by another 5 seconds would really bug the one person that uses that watch. 
The bootloader already waits for the button to be pressed to enter fastboot, and I'd argue against adding a second wait for every bootup when booting to recovery isn't that much more involved

MagneFire:
> I'm alright withj this being a temporary solution for hoki, but I would argue against merging this on other watches where only temporary install is available.

My goal was to eventually create a recovery image that would behave identically everywhere, getting this in `hoki` would be a first step into figuring out what works and what doesn't. This just means that we should probably look into an approach that works for all platforms.

> I'd suggest the boot image checks which partition it booted from (boot or recovery) and starts the adb daemon depending on that. Or we just provide a separate recovery image which always boots to adb. That's how SailfishOS does it. 

Good idea!
I generally use the recovery partition on `hoki` to boot into AsteroidOS and use the boot partition to boot into WearOS, but this has the side effect of WearOS restoring the recovery partition upon boot. So my approach isn't great anyways.

There's no real disadvantage other than figuring out where the image is booted from right? Because this sounds like a good approach that also has the benefit of working similarly to how Android and its recovery partition works.

> the issue for me is that bootup on koi (casio wsd-f20, which also only uses 'temporary' install) is already incredibly slow and slowing it down by another 5 seconds would really bug the one person that uses that watch. The bootloader already waits for the button to be pressed to enter fastboot, and I'd argue against adding a second wait for every bootup when booting to recovery isn't that much more involved

We could also reduce this delay to maybe one second, which is the boot logo animation time. That way we can instruct people to press a button or do some other action during this time.
Additionally we could think about other combinations, keep holding the touch screen, or keep holding an auxiliary button to access adb from the boot image.

----

On another note, I did notice that when you start `adbd` in the recovery image, then kill it to continue booting `adb` will not work until the next boot, I'm guessing @beroset you've already observed this too?

meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
beroset: I was wondering if this was common knowledge or not.  I'll document better.

beroset:
The issue I was trying to address was that once one has done a temporary install on `hoki`, there seemed to be no convenient way (and definitely no documented way) to install a new OS image.  You'd have to either install a boot image that always explicitly waits, and then reboot into recovery mode again, or wipe the whole watch and start again, either of which takes far longer than 5 seconds.

The alternative of killing `adbd` to continue, as @MagneFire observed, results in ADB not working until the next boot, which is annoying and, to my knowledge, unexplained behavior, so I'm not convinced that a separate recovery image would solve the original problem in a satisfactory way.

As always, I appreciate your collective thoughts on how to better solve this.

MagneFire:
> The issue I was trying to address was that once one has done a temporary install on `hoki`, there seemed to be no convenient way (and definitely no documented way) to install a new OS image. You'd have to either install a boot image that always explicitly waits, and then reboot into recovery mode again, or wipe the whole watch and start again, either of which takes far longer than 5 seconds.
> 

What's your opinion on reducing this to let's say 1 second or so? The user already has to 'spam' the button to get it ti enable adbd anyways. This might also be good enough for other platforms?

> The alternative of killing `adbd` to continue, as @MagneFire observed, results in ADB not working until the next boot, which is annoying and, to my knowledge, unexplained behavior, so I'm not convinced that a separate recovery image would solve the original problem in a satisfactory way.
> 

Yup, this is unexplained, can also reproduce this in `beluga`.

> As always, I appreciate your collective thoughts on how to better solve this.

meta-hoki/recipes-core/initrdscripts/initramfs-scripts-android/init.machine.sh
```diff
@@ -17,4 +17,34 @@ if [ ! -e $BOOT_DIR/etc/asteroid/machine.conf ] ; then
...
+
+# wait for the middle button to be pressed, and if it is,
+# kill the calling process group with SIGUSR1
+# The format of these events is documented here:
+# https://github.com/torvalds/linux/blob/master/include/uapi/linux/input.h#L28-L47
+# u32 - tv_sec
+# u32 - tv_usec
+# u16 - type
+# u16 - code
+# s32 - value
```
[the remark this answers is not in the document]
beroset: Fixed!

MagneFire (approved):

MagneFire:
@beroset Totally forgot, but we already have a 1 second delay in our boot process: https://github.com/AsteroidOS/meta-asteroid/blob/aba31eee7bac4f4c97669456326383a5eef3c7f6/recipes-core/initrdscripts/initramfs-scripts-android/init.sh#L142

The downside of removing the `sleep 1` in this PR has that we implicitly rely on the delay in the general init script. But this could be a reason to move some of this logic to the `init.sh` script too?

PR #260, opened 2025-02-16
WIP: initrdscripts: Fix init.machine FILES variable pollution

dv1:
Several initrd scripts pollute FILES:${PN} by adding init.machine, which ends up masking errors in other smartwatch specific bbappends, specifically that they are missing their own init.machine addition. This means that if only one smartwatch layer is used, this pollution can cause build errors because init.machine is not added properly.

Marked as WIP because up until now, other layers may have inadvertedly relied on that pollution, so this commit may cause a build failure due to the now-missing init.machine inclusion. The proper fix is to add `FILES:${PN}:append:<machine name> = "/init.machine"` lines to all layers.

### DIFF
--- a/meta-lenok/recipes-core/initrdscripts/initramfs-scripts-android_1.0.bbappend
+++ b/meta-lenok/recipes-core/initrdscripts/initramfs-scripts-android_1.0.bbappend
@@ -7,4 +7,4 @@ do_install:append:lenok() {
     install -m 0755 ${WORKDIR}/init.machine.sh ${D}/init.machine
 }
 
-FILES:${PN} += "/init.machine"
+FILES:${PN}:append:lenok = "/init.machine"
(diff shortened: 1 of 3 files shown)


MagneFire:
> These look good to me, but I think we are missing some.  
> ```
> grep -ir 'FILES:${PN} ' *
> meta-hoki/recipes-android/android/android-system-data_hoki-p.bb:FILES:${PN} = "/system /vendor"
> meta-mtk6580/recipes-core/mtk6580-scatter/mtk6580-scatter_1.0.bb:FILES:${PN} += "/boot/MT6580_AsteroidOS_scatter.txt"
> meta-narwhal/recipes-android/android/android-audio-policy_narwhal.bb:FILES:${PN} = "${sysconfdir}/pulse/"
> meta-narwhal/recipes-core/hands-timesync/hands-timesync.bb:FILES:${PN} += "${systemd_system_unitdir} /etc/systemd/"
> meta-ray/recipes-android/android/android-system-data_ray.bb:FILES:${PN} = "${sysconfdir}/udev /system /vendor"
> meta-rinato/recipes-kernel/samsung-smartwatch-firmware/samsung-smartwatch-firmware_git.bb:FILES:${PN} += " /usr/lib/firmware "
> meta-skipjack/recipes-android/android/android-system-data_skipjack-p.bb:FILES:${PN} = "/system /vendor"
> ```


Good point, I did see those as well, but didn't do them because they are not `bbappend` files. These recipes are only used by the relevant `MACHINE`. Others don't use them.

PR #262, opened 2025-02-23
triggerfish: Include hoki initramfs patch

Avellea:
Patch and direction by @MagneFire, this gets `triggerfish` to a booting (albeit poor) state. Tested with a Fossil 5 Julianna.


MagneFire:
Awesome thank you!

PR #263, opened 2025-02-27
triggerfish: Add basic wifi support

ZTube:
This commit adds basic wifi functionality to the fossil gen 5 triggerfish in order for configuration via connman / wpa_supplicant to work.

To do that firmware blobs are now mounted at the correct location where they can be found via firmwared.
An external kernel module 'wlan' to provide the interface is included and loaded at startup

Enjoy :D


MagneFire (approved):
Awesome thank you!!

PR #275, opened 2025-11-10
Add geoclue to each appropriate watch

beroset:
For watches that have geolocation hardware, this adds geoclue to the machine configuration.


beroset:
I have rethought this and have decided to close the PR.  For the few apps that need it, `geoclue` can simply be listed as a dependency.

PR #281, opened 2026-02-06
Fix package installation to /usr/bin, not /bin

beroset:
Since the usr merge in 2012 (see https://fedoraproject.org/wiki/Features/UsrMove), all Linux systems should install to /usr/bin rather than /bin which is now just a symlink to /usr/bin.  This corrects a few scripts to make sure they pass Yocto qa tests.

### DIFF
--- a/meta-catfish/recipes-android/udev-droid-system/udev-droid-system_1.0.bb
+++ b/meta-catfish/recipes-android/udev-droid-system/udev-droid-system_1.0.bb
@@ -9,6 +9,6 @@ UNPACKDIR = "${S}"
 B = "${S}"
 
 do_install:append() {
-    install -d ${D}/lib/udev/rules.d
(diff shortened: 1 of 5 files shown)


meta-catfish/recipes-android/udev-droid-system/udev-droid-system_1.0.bb
```diff
@@ -9,6 +9,6 @@ UNPACKDIR = "${S}"
 B = "${S}"
 
 do_install:append() {
-    install -d ${D}/lib/udev/rules.d
-    install -m 644 ${UNPACKDIR}/998-droid-system.rules ${D}/lib/udev/rules.d/
+    install -d ${D}/usr/lib/udev/rules.d
```
MagneFire: Wouldn't it make more sense to use `${libdir}` instead of hardcoding `/usr/lib` everywhere?

PR #283, opened 2026-02-08
Add support for `rubyfish`

MagneFire:
This work supersedes https://github.com/AsteroidOS/meta-smartwatch/pull/227.

It  depends on https://github.com/AsteroidOS/lcd-tools/pull/9

Support the amount of things working is mostly the same as the previous work. This contains the following changes:
- Address comments with regards to used device trees
- Cleanup kernel patches, files and patch name order
- Fix LCD
- Fix rover
- Remove speaker machine config (audio is broken)
- Removed work that doesn't work or hasn't been tested (like audio, nfc)
- Cleanup android-init services

### DIFF
--- a/meta-rubyfish/conf/machine/rubyfish.conf
+++ b/meta-rubyfish/conf/machine/rubyfish.conf
@@ -17,4 +17,4 @@ PREFERRED_VERSION_android = "armv7+pie64"
 PREFERRED_PROVIDER_virtual/kernel = "linux-rubyfish"
 PREFERRED_VERSION_linux = "4.9+pie"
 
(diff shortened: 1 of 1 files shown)


moWerk (approved):
Tested on my rubyfish and found all claimed features working.
One remaining issue is very low brightness after first boot and thus in firstbootconfig. Toggeling brightness in the quickPanel seems to solve that.
The brightness is bumping very strongly with the light sensor.
To be addressed at later stage.

MagneFire:
`rubyfish` support for `lcd-tools` has been merged and is already available in main via https://github.com/AsteroidOS/meta-asteroid/pull/223

PR #301, opened 2026-04-15
rubyfish: mce: Provide custom brightness tune

MagneFire:
The minimum brightness on rubyfish is too low, resulting in in a completely unreadable screen.
This bumps the minimum brightness to 28% which seems to improve it.

draft as I want to test it for a little while


FlorentRevest (approved):

moWerk (commented):
Can confirm that 28% solves the issue. But in my testing i start to see the display barely at 5% and from 10% on its quite noticable even in daylight. We had users asking in chat for low display brightness for use in the night and i wonder if 28% would not be a bit to bright for them.
Maybe 15-20% could be a middleground?

MagneFire:
> Can confirm that 28% solves the issue. But in my testing i start to see the display barely at 5% and from 10% on its quite noticable even in daylight. We had users asking in chat for low display brightness for use in the night and i wonder if 28% would not be a bit to bright for them. Maybe 15-20% could be a middleground?

Now adjusted to 17%, can confirm this works well too.

moWerk (approved):
Tested 17% as well. lgtm

PR #304, opened 2026-04-26
rubyfish: Fix WiFi saved networks

MagneFire:
~~The bcmhd WiFi driver reads the /persist/wifi_nv.bin file to obtain the device specific MAC address.
This driver is loaded during kernel boot, which results in it failing to read the MAC address file as the /persist partition isn't yet mounted. It then generates a random MAC address which will also be used as the adapter identifier for the rest of the boot.~~

~~Once userspace attaches the adapter, the MAC address file is read again, which results in the correct MAC address being used for the interface. But as the wrong identifier is used, the system 'forgets' the connected WiFi network every boot.~~

~~To work around this issue, fallback to a hardcoded MAC address.~~

This changes how the BCMDHD WiFi driver is loaded. Instead of it being built-in compile it as a module and load it later during the boot process.
This fixes an issue where the driver would generate a random MAC address used for identification resulting in it forgetting previously connected networks.


KonTy:
this seems like a band-aid and a dirty fix, we shouldn't be hard coding macs into the kernel. The better way is to unload the driver completely and when you toggle the button it will reload thedriver which will see the bin file and get the right  address. I fixed it in my fork, but it is so far apart from this codebase I don't think we can do it that way here.  
I think is asteroidos code we could blacklist the wifi driver so we control the order and move it to /etc/modprobe.d/blacklist-bcmhd.conf containing:
blacklist bcmhd

and then do a oneshot Systemd service (load-wifi-driver.service) that explicitly loads the module, but strictly after the local filesystems are mounted and before the network manager (ConnMan) starts:

IMHO this is better? well both of these approaches guarantee that the driver never initializes until /persist is successfully mounted, meaning it reads the correct MAC address on the very first try, preventing ConnMan from ever seeing (and saving) a ghost MAC address

PS try my fork, I'm doing a lot of battery optimizations and perf optimizations, I got a battery drain to less then 1% per hour on OPPO Watch, still trying to make it better.

MagneFire:
> this seems like a band-aid and a dirty fix, we shouldn't be hard coding macs into the kernel. The better way is to unload the driver completely and when you toggle the button it will reload thedriver which will see the bin file and get the right address. I fixed it in my fork, but it is so far apart from this codebase I don't think we can do it that way here. I think is asteroidos code we could blacklist the wifi driver so we control the order and move it to /etc/modprobe.d/blacklist-bcmhd.conf containing: blacklist bcmhd
> 

Totally agree that this is a band-aid fix. But given that it's only used as an identifier, I think it's fine (later during the boot the correct mac address is loaded).
I did think about unloading the driver, but that's not possible as the kernel is not compiled as a module, instead it's baked in.
I find it weird that the kernel driver is even compiled in the first place as it's not enabled in the kernel configuration: https://github.com/AsteroidOS/meta-smartwatch/blob/91d7fbfd9de6bbcf01ae23793fdf91d619d48265/meta-rubyfish/recipes-kernel/linux/linux-rubyfish/defconfig#L1600

> and then do a oneshot Systemd service (load-wifi-driver.service) that explicitly loads the module, but strictly after the local filesystems are mounted and before the network manager (ConnMan) starts:
> 
> IMHO this is better? well both of these approaches guarantee that the driver never initializes until /persist is successfully mounted, meaning it reads the correct MAC address on the very first try, preventing ConnMan from ever seeing (and saving) a ghost MAC address
> 

Yes, I do think it's better. I'll do some further research. Thank you for your review 😄 

> PS try my fork, I'm doing a lot of battery optimizations and perf optimizations, I got a battery drain to less then 1% per hour on OPPO Watch, still trying to make it better.

FlorentRevest (approved):

PR #306, opened 2026-05-27
skipjack: linux-skipjack: touchscreen: focaltech: Override size for t unny

### LINKED ISSUE #150: Offer two seperate images for skipjack and tunny
As recently discovered, the TicWatch E2 does work with the S2 images.
Both need a fix for the touch input when using the skipjack images.

MagneFire:
tunny and skipjack share the same specs other than the display resolution (400x400 vs 360x360).

To avoid building for two platforms dynamically figure out what the touchscreen size should be based on the kernel command line.

This fixes #150 

Marked as draft as this has not been tested.


FlorentRevest (approved):

moWerk (approved):
Tested on two tunny watches (S2/E2) and skipjack. The fix works on tunny without regressing on skipjack. Great work.

PR #311, opened 2026-06-10
Pixel Watch 2 (aurora) port

FlorentRevest:
This is our first GKI port, our first 5.15 kernel port, our first Halium port and Android 13 port so there's quite a lot going on here - see the commits for details.

What I tested to work:
- Display
- Touch
- Bluetooth
- Haptics
- Sensors (I only tested compass and HRM but I assume the rest works too)
- AoD
- Speaker
- USB (I only tested adb so it's hardcoded in the port - I have no way to test RNDIS right now)
- WLAN

This runs butter smooth on a gorgeous screen and I love it :)

This depends on:

- https://github.com/AsteroidOS/meta-asteroid/pull/238
- https://github.com/AsteroidOS/meta-asteroid/pull/248
- https://github.com/AsteroidOS/asteroid/pull/344

To run this you first need to flash `TWD9.240405.001, Apr 2024` from https://developers.google.com/android/images-watch. Making the port work with newer versions is left as an exercise to the reader. :) 

Then you can flash with:

```
fastboot set_active b
fastboot flash userdata asteroid-image-aurora.rootfs.ext4
fastboot flash boot boot.img
fastboot flash vendor_kernel_boot vendor_kernel_boot.img
fastboot flash init_boot init_boot.img
fastboot reboot
```

Big up to UBPorts for their inspiration https://gitlab.com/ubports/porting/community-ports/android13/google-eos/google-eos and to Claude for crunching through a lot of the tedious bits.

Hopefully this should make a Google Pixel Watch 3 port a breeze since they share the same SoC and I assume a lot would be exactly the same


moWerk:
Initial report back, i had two issues when building.

Firstly, got this error
````
$ bitbake asteroid-image Loading cache: 100% |                                                                                                                                                        | ETA:  --:--:-- Loaded 0 entries from dependency cache. ERROR: ExpansionError during parsing /home/mo/Git/asteroid/src/meta-smartwatch/meta-rinato/recipes-kernel/linux/linux-rinato/linux-rinato_next.bb                             | ETA:  bb.data_smart.ExpansionError: Failure expanding variable LINUX_MINOR, expression was ${@(d.getVar('LINUX_VERSION') or "x.y").split('.')[1]} which triggered exception IndexError: list index out of range The variable dependency chain for the failure is: LINUX_MINOR ERROR: Parsing halted due to errors, see error messages above
````

Which i worked around with editing `meta-rinato/recipes-kernel/linux/linux-rinato/linux-rinato_next.bb` and changing `LINUX_VERSION ?= "next"`  to `LINUX_VERSION ?= "next.0"`
 
 The second issue is that the vkb was not correcly stripped of debug_info and ended up being 150MB.
 The fix for me editing then bblayer.conf and adding a strip step using llvm-strip
 
 ```
  # ─── Step 2.5: strip debug info from kernel modules ───     # Modules are aarch64 ELF but this recipe runs in an armv7 context.     # The default strip cannot handle them. llvm-strip from clang-native     # is architecture-agnostic and handles any ELF target.     LLVM_STRIP=$(find ${STAGING_BINDIR_NATIVE} -name "llvm-strip" | head -1)     if [ -z "$LLVM_STRIP" ]; then         bbfatal "llvm-strip not found in ${STAGING_BINDIR_NATIVE}"     fi     find ${WORKDIR}/vkb_ramdisk -name "*.ko" -exec "$LLVM_STRIP" --strip-debug {} \;     bbnote "stripped debug info from aarch64 kernel modules" 
 ```
 
 Claude analyzing the .ko files:

The modules in `vkb_ramdisk` are not stripped — `file` confirms
`with debug_info, not stripped` on all `.ko` files. `wlan.ko` alone
is 320MB unstripped.

Root cause: `aurora-boot-images.bb` runs in the armv7 recipe context
but the kernel modules are aarch64 ELF. The default strip tool in that
context cannot handle them and silently does nothing.

Fix: `llvm-strip --strip-debug` from `clang-native` is
architecture-agnostic and works correctly. A strip pass on the
assembled `vkb_ramdisk` before the cpio step reduces the result:

- `wlan.ko`: 320MB → 11MB
- full ramdisk lz4: 150MB → 7.5MB (well within the 64MB cap)

Proposed addition to `do_compile` in `aurora-boot-images.bb` between
the python assembly step and the cpio step, with `clang-native` added
to DEPENDS.

moWerk:
Ok, finally got it. And surprise. The vbmeta was not my issue. Even after trying 

`fastboot flash vbmeta --disable-verity --disable-verification vbmeta.img`
and
`fastboot flash vbmeta_system --disable-verity --disable-verification vbmeta_system.img`
The watch just reboot.
But when trying slot_a, it suddenly worked. 

`fastboot set_active a
fastboot flash boot_a boot.img
fastboot flash vendor_kernel_boot_a vendor_kernel_boot.img
fastboot flash init_boot_a init_boot.img
fastboot flash userdata asteroid-image-aurora.rootfs.ext4
fastboot reboot`

But now i can not really say if both was necessary.

FlorentRevest:
Ah yeah I guess it should be whatever slot android flashed to 🤔 in my case it was b but in your case a

moWerk:
I used the flash_all.sh in the TWD9 firmaware to "downgrade". 
And afaiu from my past phone flashing endevours, selecting a slot is an edge case and usually the slot wearos was flashed to should be selcted once it was booted successfully. Making the slot selection unnecessary.
Maybe an instruction "Boot to wearos at least once before flashing AsteroidOS" would do?

PR #313, opened 2026-06-18
nemo: kernel bring-up cleanup and bass-parity patches

moWerk:
This is the first round of cleanup for resuming the nemo (LG Watch Urbane 2nd gen) port, abandoned ~4 years ago after initial bring-up (display, touch, USB) landed with help from MagneFire and dodoradio.

Marked as draft, this is groundwork, not a complete bring-up. Opening early for visibility and feedback on direction before more commits land.

**Edit:** force-pushed a clean replay of this branch to address review feedback — each new patch file now lands in the same commit as its `.bb` insertion line (previously these were decoupled, breaking bisectability per @FlorentRevest's comment), and the nitrous bluesleep stub's patch header now follows the same `From:`/`Date:`/`Subject:` convention as the rest of the series instead of the ad-hoc two-liner it had. No functional changes from that push beyond those two fixes; diffed the full tree against the prior branch state to confirm.

**Update:** Bluetooth is now working. The patchram service file was an untouched copy of bass's unit, pointing at firmware for a BCM4343A0 chip; nemo's actual hardware is a BCM43430A1, confirmed via a WearOS factory dump. The firmware file for that chip didn't exist on device, so the patchram load silently failed on every boot and the chip never came up far enough to respond on the UART at all. Shipping the correct firmware fixed it. Also wired the existing `bluesleep_hci_event` stub into nitrous's wake mechanism while testing, mirroring how bass's `bcm_bluesleep` driver uses the same hook.

**What's in this PR right now**

* Kernel patches 0003-0005 renumbered (pure rename, no content change) to make room for new additions
* Three patches ported from meta-bass for MSM8226 SoC parity: mainline 4.1 Bluetooth subsystem backport, bluesleep fix, vibrator API conversion
* getrandom() syscall backport, required for glibc 2.41 compatibility on this 3.10 kernel. This fixed SSH-over-USB, which was silently broken without it.
* nitrous bluesleep stub now wired into the existing wake mechanism rather than a no-op
* kernel recipe updated to wire up all renumbered and new patches, verified to apply cleanly on hardware
* WiFi firmware wired into a `do_install:append` in the android-system recipe — correct `_a1` variants for the BCM43430_A1 chip, extracted from the WearOS factory dump, installed over the stock tarball's wrong generic symlinks
* Bluetooth firmware fixed the same way — correct `BCM43430A1.hcd` for the actual chip, wired into the `brcm-patchram-plus` recipe via `SRC_URI`/`FILES`

**Confirmed working on hardware (LG Watch Urbane 2nd edition / nemo)**

* Boot, display, touch, USB (carried over from original bring-up)
* SSH over USB (fixed this round, was broken)
* WiFi: associates and gets a real IP on first cold boot, no manual firmware copy required
* Bluetooth: pairs and connects to Gadgetbridge, confirmed via BLE pairing flow

**Still open, more commits coming to this PR**

* Bluetooth pairing requires manually entering a PIN rather than auto-trusting. A similar trust issue came up on an earlier port and was fixed by MagneFire at the time; investigating whether the same fix applies here
* AOD/LPM: display does not enter true low-power state, visible brightness pulsing every few seconds. Unaffected by the bluesleep wake-mechanism fix above
* Sensors: ADSP PIL fails to load with invalid firmware metadata, not yet started
* Power off and reboot from software do not work, screen stays black and device keeps draining; fastboot power-cycle works as a workaround. Possible watchdog or PMIC sequencing issue, not yet investigated
* connman does not auto-reconnect to a saved WiFi network after reboot, requires manual reconnect. Possibly not nemo-specific, same behavior observed on rubyfish

Will keep this description updated as commits land. Undrafting once the above list is resolved or scoped down and initial review feedback is addressed.


meta-nemo/recipes-kernel/linux/linux-nemo_lp-mr1.bb
```diff
@@ -11,9 +11,15 @@ COMPATIBLE_MACHINE = "nemo"
...
-    file://0005-vfs-allow-umount-to-handle-mountpoints-without-reval.patch \
+    file://0003-Backport-mainline-4.1-Bluetooth-subsystem.patch;striplevel=1 \
+    file://0004-bluesleep-Use-kernel-s-HCI-events-instead-of-proc-bl.patch \
+    file://0005-msm_pwm_vibrator-Convert-timed_output-APIs-to-ff_mem.patch \
+    file://0006-ARM-uaccess-remove-put_user-code-duplication.patch \
+    file://0007-random-introduce-getrandom-2-system-call.patch \
+    file://0008-ARM-wire-up-getrandom-syscall.patch \
+    file://0009-ARM-8933-1-replace-Sun-Solaris-style-flag-on-section.patch \
+    file://0010-vfs-allow-umount-to-handle-mountpoints-without-reval.patch \
+    file://0011-nemo-nitrous-bluesleep-stub.patch \
```
FlorentRevest, line 22: It's best to do each of those insertions when the patches actually get introduced. This helps with reverting (eg: if we only want to revert the nitrous change, we would have to manually split up this patch) and with bisectability (eg: if we git checkout one commit before this, the recipe would fail to build because it refers to 0003-ARM-uaccess-remove-put_user-code-duplication.patch which no longer exists.

meta-nemo/recipes-kernel/linux/linux-nemo/0011-nemo-nitrous-bluesleep-stub.patch
```diff
@@ -0,0 +1,18 @@
+From: nemo port
```
FlorentRevest: This is a weird patch header formatting - have a look at other patches for examples of what this should look like

FlorentRevest (commented):
overall looks good!

meta-nemo/recipes-kernel/linux/linux-nemo_lp-mr1.bb
moWerk: Moved the patch additions to the respective commits where the patch is added.

meta-nemo/recipes-android/android/files/adsp.b00
MagneFire, line 1: Please avoid adding binary files to git repo's. Then there's also the source of these files: How does the license work for them? Generally these firmware files are mounted (https://github.com/AsteroidOS/meta-smartwatch/blob/master/meta-narwhal/recipes-core/base-files/base-files/fstab#L6) why doesn't that work here?

meta-nemo/recipes-android/android/android_nemo-mm.bb
```diff
@@ -2,14 +2,33 @@ inherit gettext
 
 SUMMARY = "Downloads the LG Watch Urbane /system and /usr/include/android folders and installs them for libhybris"
 LICENSE = "CLOSED"
-SRC_URI = "https://dl.dropboxusercontent.com/s/hta18rrkn7wenc9/system-M1D64S.tar.gz"
+SRC_URI = "https://dl.dropboxusercontent.com/s/hta18rrkn7wenc9/system-M1D64S.tar.gz \
```
MagneFire, line 5: Seeing this now. It's probably better to update the system archive and add these additional files there if mounting the firmware partition isn't enough.

MagneFire (changes requested):
Great work!

Generally looks good, but please don't add binary files to a git repo.

moWerk:
superceeded by #317

PR #315, opened 2026-06-21
Rinato: hack kernel version to supress errors.

dodoradio:
yocto seems to expect KERNEL_VERSION to be in x.y format.


MagneFire (approved):

dodoradio:
apparently this isn't universal - kido hasn't been able to reproduce it. But it does cause issues for containerised builds, as reproduced by both @moWerk and @beroset

moWerk (approved):
Can confirm that it is necessary for all wathces i tested last week..When i mentioned this next.0 fix in chat, Kido noted he would be ok with it given an appropriate explanatory comment.

FlorentRevest:
> Kido noted he would be ok with it given an appropriate explanatory comment.

Ahah, which was forgotten

PR #316, opened 2026-06-24
nemo: Nougat-based Android userspace

moWerk:
This branch continues the LG Watch Urbane 2nd Edition (nemo) port, replacing the Marshmallow Android userspace base with one built from the genuine nemo WearOS factory dump.

### Why a new base

Investigation during the previous session revealed that `android_nemo-mm.bb` fetched `system-M1D64S.tar.gz` — LG/bass's actual Android 6.0.1 (Marshmallow) factory dump (`lge/bass/bass:6.0.1/M1D64S/3505725:user/release-keys`), not nemo's. The real nemo WearOS dump is Android 7.1.1 (`NWD1.180306.004`, SDK 25), confirmed against the latest OTA on a live paired device. This mismatch was the leading hypothesis for sensors remaining broken after every other file-level fix was exhausted.

This branch builds on the kernel and Bluetooth firmware work from `nemo-port-polish` and adds a correct Nougat system-dir tarball.

### New commits on this branch

**`nemo: switch android userspace to Nougat base`**
Introduces `android_nemo-ng.bb` targeting a system-dir tarball built from the real nemo WearOS factory dump. HAL components (hwcomposer, gralloc, memtrack, servicemanager, logd) recompiled against `github.com/MagneFire/android_manifest` branch `nougat-mr1.8-release`. The tarball is hosted externally (mowerk.net); no binaries in the repository. Adds `/system` and `/vendor` root symlinks required by the Android bionic linker. Patches `hwcomposer.msm8226.so` with `patchelf --add-needed libEGL_adreno.so` to satisfy a symbol dependency (`eglGpuPerfHintQCOM`) that the bionic linker must resolve before constructors run.

**`nemo: create /dev/graphics/fb0 and start Android binder servicemanager`**
`hwcomposer.msm8226.so` opens `/dev/graphics/fb0` at init; this path does not exist on AsteroidOS. A `tmpfiles.d` rule creates it as a symlink to `/dev/fb0`. A new system unit starts `servicemanager` at `multi-user.target`; the user session (and thus asteroid-launcher) starts after the system reaches that target, so servicemanager is up before the HAL is opened. There is no explicit `After=` dependency in the launcher unit — the ordering relies on the system/user-session boundary.

**`nemo: remove cont-splash and ulps from panel DTS`**
Removes `qcom,cont-splash-enabled` and `qcom,ulps-enabled` from both panel nodes. With these absent the kernel reports "Continuous splash disabled" and "ulps feature disabled"; the display initialises cleanly. The specific failure modes when these properties are set were not traced; removed empirically.

**`nemo: update system-dir build instructions for Nougat`**
Updates `README-system-dir` to document the Nougat build process: correct Android version (7.1.1), correct manifest (`MagneFire/android_manifest -b nougat-mr1.8-release`), removal of the libhybris header extraction step (headers come from a separate version-pinned shared package), and fixes a typo in the final tar invocation (`czf` not `zxvf`).

### Hardware status

| Subsystem | Status |
|---|---|
| Boot | ✅ Boots to launcher automatically |
| Display | ✅ 480×480, round, renders correctly |
| Touchscreen | ✅ |
| Bluetooth | ✅ Pairs and connects to Gadgetbridge (BLE) |
| WiFi | ✅ Hardware up (wlan0 DORMANT, associates) |
| Software reboot | ✅ Now works cleanly — did not on the Marshmallow base |
| Sensors (SSC/ADSP) | ❌ sensorfwd still reports no sensors found |
| Vibration | ❌ CONFIG_INPUT_FF_MEMLESS not set in defconfig |
| Audio | ❌ No ALSA cards detected |
| BT pairing | ⚠️ Requires manual PIN entry (auto-trust not yet investigated) |
| AoD/LPM brightness | ⚠️ Pulses every ~3s instead of clean dim-and-hold |

### Scaffolding still in place

The six overrides carried from the Marshmallow recipe (WiFi firmware, ADSP firmware, `sensor_def_nemo.conf`, `sensors.msm8226.so`, `build.prop` sed, HAL symlinks) are kept intact. Each needs to be verified redundant against the Nougat base and stripped one at a time with a flash-and-verify cycle before this branch is considered clean.


moWerk:
superceeded by #317

PR #322, opened 2026-07-14
koi: add init.machine.sh for a proper install procedure

dodoradio:
this: 
- reverts a commit that broke usb configfs
- adds init.machine.sh and e2fsprogs

Currently, flashing koi is a bit of a mess, as android seems to sometimes brick this watch. the current install instructions rely on rebooting to android after unlocking the bootloader, which feels risky.
These changes allows for a safer install process for koi, which would roughly involve:
- unlocking the bootloader
- flashing the asteroidos boot image to the boot partition
  - optionally flashing asteroidos boot image to recovery partition, so there's no risk the watch accidentally boots into android and corrupts itself
- booting the watch and waiting for it to show up on usb as `Casio WSD-F20 INITRD`
  - reformatting the userdata partition with `adb shell mkfs.ext4 /dev/mmcblk0p3 -O^metadata_csum` (the bootloader seems to sometimes mess this up)
  - rebooting the watch with `adb reboot` (to allow the sdcard partition to be remounted. this could also be achieved with `adb shell mount /dev/mmcblk0p3 /sdcard`)
- after watch reboots and again shows up as `Casio WSD-F20 INITRD`, push the rootfs with `adb push asteroid-image-koi.rootfs.ext4 /sdcard/asteroidos.ext4`
- finally reboot the watch into asteroidos with `adb reboot`

For some reason, psplash fails to show anything onscreen when the watch is in the initrd. On normal boot, something similar happens, where it takes a while for the splash screen to show up. This seems to not be improved by the addition of msm-fb-refresher.


MagneFire (approved):
Looks good, thank you!

PR #325, opened 2026-07-18
Medaka: modularise wifi driver so it can have a persistent mac address

dodoradio:
Currently, medaka's [wifi platform driver](android.googlesource.com/kernel/exynos/+/refs/heads/android-exynos-medaka-4.4-pie-wear-dr/drivers/misc/casio/board-medaka-wifi.c) fails to load a mac address from the fs, and ends up generating a random one every time. 

To fix this, I've built the driver as a module, so it can load after the rootfs is mounted. There's a pair of kernel patches, which hack around issues that the vendor wifi driver caused when built as a module.

Once fixed, the driver looks for a /persist, and I can't find anything like that on this watch (usually /persist would be a partition, but there isn't one named that way). I have no idea how the mac address was loaded under the original wearos. So instead I've put a script into android-init which generates a mac address at the right path if there isn't one. Maybe this needs to eventually be its own recipe, like msm-wifi-enabler.


meta-medaka/recipes-android/android-init/android-init/macaddr-generator
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# Medaka expects a wifi mac address in /persist, but doesn't actually have a persist partition
```
FlorentRevest: On aurora for example I have

```
sh-5.2# mount | grep "persist"
/dev/mmcblk0p16 on /persist type ext4 (rw,nosuid,nodev,noatime)
/dev/mmcblk0p16 on /var/lib/lxc/android/rootfs/mnt/vendor/persist type ext4 (rw,nosuid,nodev,noatime)
sh-5.2# ls /persist/
battery  display   hvdcp_opti  magisk  sensors	       vpp
crown	 haptics   iar_db      rfs     system.journal
data	 hlos_rfs  lost+found  secnvm  time
sh-5.2#
```

have you tried mounting various /dev/mmcblk* and looking for a wifi directory in them ?

meta-medaka/recipes-android/android-init/android-init/macaddr-generator
```diff
@@ -0,0 +1,14 @@
...
+# put a file in persist in the rootfs anyway
+
+MAC_FILE="/persist/wifi/.macaddr"
+DIR_PATH="/persist/wifi"
+
+if [ ! -f "$MAC_FILE" ]; then
+    mkdir -p /persist/wifi/
```
FlorentRevest: Maybe use dirname $MAC_FILE and drop DIR_PATH ?

meta-medaka/recipes-android/android-init/android-init/macaddr-generator
dodoradio: yes. I have looked through all the partitions, and I've spent ages grepping through vendor system and oem in search of any references to persist. I'm pretty confident in saying there are none.

dodoradio: there's a few raw data partitions which seem like they'd provide the mac address, but they either contain a null mac, or they are the same across two separate devices

meta-medaka/recipes-android/android-init/android-init/macaddr-generator
dodoradio: ok, I'll try that

meta-medaka/recipes-kernel/linux/linux-medaka/0003-bcmdhd-always-load-firmware-from-BCMDHD_FW_PATH.patch
```diff
@@ -0,0 +1,26 @@
...
+
+diff --git a/drivers/net/wireless/bcmdhd/dhd_linux.c b/drivers/net/wireless/bcmdhd/dhd_linux.c
+index a9091bf772fe..bd2d2008c204 100644
+--- a/drivers/net/wireless/bcmdhd/dhd_linux.c
++++ b/drivers/net/wireless/bcmdhd/dhd_linux.c
+@@ -9626,7 +9626,7 @@ bool dhd_update_fw_nv_path(dhd_info_t *dhdinfo)
+ 	 */
+ 
+ 	/* set default firmware and nvram path for built-in type driver */
+-	if (!dhd_download_fw_on_driverload) {
```
FlorentRevest, line 19: Looking at the driver https://android.googlesource.com/kernel/exynos/+/refs/heads/android-exynos-medaka-4.4-pie-wear-dr/drivers/net/wireless/bcmdhd/dhd_linux.c it looks like there's a lot more stuff that depend on dhd_download_fw_on_driverload and maybe flipping one branch instead of the other can cause all sorts of unexpected issues. Maybe a safer patch would be

```
  /* Flag to indicate if we should download firmware on driver load */
- uint dhd_download_fw_on_driverload = TRUE;
+ uint dhd_download_fw_on_driverload = FALSE;
```

Is this something you've tried ? Some of the logic that would be change would be around wakelocks for example so I would fear that the patch as it is right now could cause battery issues that will be hell to debug or so.

meta-medaka/recipes-android/android-init/android-init/macaddr-generator
dodoradio: thanks for this solution! I've implemented it now

meta-medaka/recipes-kernel/linux/linux-medaka/0003-bcmdhd-always-load-firmware-from-BCMDHD_FW_PATH.patch
dodoradio: yeah, I looked at the other things that are conditional on this and intentionally chose to just flip this one case. I'm not trying to fiddle with any of the firmware loading logic, I just want to fix the firmware path, so I decided it was safer than flipping the value globally.

FlorentRevest: Mh ok - then just keep it in mind in case issues that could be related pop up

PR #329, opened 2026-07-25
underclock: drop per-device recipes in favour of the generic one

FlorentRevest:
Every machine carried its own recipes-core/underclock recipe whose script differed only in device-tuned values. meta-asteroid now ships a single generic, machine-independent underclock recipe that guards each sysfs write behind an existence check, so it runs correctly on any watch.

Remove all 26 per-device copies. No machine.conf changes are needed: they all keep 'underclock' in IMAGE_INSTALL, which now resolves to the generic recipe. tetra's PowerHAL specifics (interactive governor and the kona-fb / interactive tunable chowns) are folded into the generic script, gated on the tetra-only kona-fb platform device, so tetra no longer needs a bespoke recipe either.


MagneFire:
This will reduce responsiveness on at least `beluga` and `catfish` where the GPU isn't locked to their lowest frequency. On those platforms using different frequencies are allowed and it has been shown that this has improved responsiveness (higher FPS) on those. 

Edit: I see we do a similar thing for `sturgeon`.

FlorentRevest:
We never really tested whether these GPU frequencies actually affected battery life for better or worse. Back in the early dory days I just intuited that a lower frequency probably meant better battery life but at some point I was also told that lower frequency means more time with the GPU-CPU bus open and generally speaking more time with the SoC cluster being on which might actually be detrimental to the overall battery life.

Maybe it's finally time we actually measure whether these GPU frequency changes help in any meaningful way the overall battery life and if not then dropping it entirely from the underclock script.

I feel that's something where we would benefit from @moWerk's fish tank too

PR #332, opened 2026-07-27
Improve boot.img generation (less bbclasses - better recipe names)

FlorentRevest:


dodoradio:
The code changes look good. The recipe renaming will make these recipes easier to find. Does it all still build fine?

PR #334, opened 2026-08-05
Move to Yocto whinlatter

MagneFire:
Even though whinlatter is already eol I think it still makes sense to move to whinlatter before moving to Wrynose as we've gotten whinlatter to work a while back.

This work supersedes https://github.com/AsteroidOS/meta-smartwatch/pull/292


beroset:
this was rebased and merged.

PR #337, opened 2026-08-16
meta-aurora: adb and ssh over one USB gadget (NCM)

moWerk:
aurora only ever offered adb, and unreliably; ssh-over-USB never worked.

Both come from usb-moded, whose udev watch dies at startup here, so it never
selects a mode. adb depended on its appsync and came up only on some boots, and
developer_mode asks for rndis, which these kernels do not provide
(CONFIG_USB_CONFIGFS_RNDIS unset).

This links ncm.0 alongside ffs.adb: one composite gadget carrying adb and
network at the same time, so nothing switches between them. sshd needs no
change. A second commit fixes the gadget serial, which was being taken from the
wrong place and made `adb -s <serial>` fail against a healthy watch.

Worth noting for testing: RNDIS has never been usable on macOS, while CDC-NCM is
supported there natively - kido wanted to try exactly this.

Verified on aurora across reboots: adb and ssh both up unaided, correct serial,
and a link-local that is identical on every boot (the NCM MAC is derived from
the device tree DSN), so a host reaches the watch at fe80::...%iface with no
address assignment and no root.

Only meta-aurora is touched. The same approach applies to sol (its kernel has
CONFIG_USB_CONFIGFS_NCM=y and no rndis) but that is a separate PR. The 4.9-era
watches (beluga, rubyfish) have no configfs gadget at all, so this does not
apply to them.

This was executed by an LLM under my direction. I have fully tested the changes and pre-reviewed the code.

### DIFF
--- a/meta-aurora/recipes-nemomobile/usb-moded/usb-moded/aurora-usb-network
+++ b/meta-aurora/recipes-nemomobile/usb-moded/usb-moded/aurora-usb-network
@@ -1,0 +1,123 @@
+#!/bin/sh
+# Add the NCM function to aurora's USB gadget, giving ssh-over-USB alongside adb.
+#
+# usb-moded's udev watch does not survive startup here, so it never selects a
+# mode: it neither starts adbd nor sets up a network function, and its
+# developer_mode asks for rndis, which these kernels do not provide. ncm.0 is
+# available, and linking it next to ffs.adb yields one composite gadget carrying
+# both channels at once.
+set -u
+
+G=/config/usb_gadget/g1
+ADDR="${AURORA_USB_ADDR:-192.168.42.2/24}"
+LOGF=/var/log/aurora-usb-network.log
+# usb-moded rewrites configs/b.1 while starting, so give it time to finish.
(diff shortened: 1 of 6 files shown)


moWerk:
Moving this to draft.

Kido's review point is the right one: the fleet needs a device-agnostic usb-moded
integration that picks whichever method a given device supports, rather than
another per-machine implementation. This works and is tested, but merging it
would add a second island (sol has an equivalent) instead of the shared layer.

Keeping it open as a reference for what the configfs-era watches need. The survey
behind it, in case it helps the general design:

- aurora, sol (5.15, configfs): ncm.0 available, no rndis, and usb-moded never
  selects a mode because its udev watch dies at startup
- beluga, rubyfish (4.9, legacy android_usb): no configfs gadget at all, all
  configfs network functions unset, and usb-moded fails to stay running - a
  different fault needing a different remedy
- detection is cheap and device-side: /config/usb_gadget vs
  /sys/class/android_usb, then "mkdir functions/ncm.test" as the capability probe

PR #339, opened 2026-08-27
sawfish: publish Python 3 in the opkg feed

monperrus:
Closes AsteroidOS/meta-asteroid#326.

Adds the existing OpenEmbedded `python3` target package to `PACKAGE_FEED` only for sawfish. AsteroidOS’s package-feed task will therefore build and index the CPython runtime and its split runtime dependencies for `armv7vehf-neon`, while the default image remains unchanged.

Validation: checked this configuration against the pinned `walnascar` OpenEmbedded Python recipe, which provides the `python3` runtime package, and ran `git diff --check`.


monperrus:
Superseded by AsteroidOS/meta-asteroid#325, which adds `python3` to the general package feed instead of scoping it to `sawfish` — thanks @MagneFire for the pointer.

Issue #224, opened 2023-09-01
[hoki] wifi support


argosphil:
TL;DR: wifi appears to work with an externally-provided driver

I tried this driver: https://github.com/Dev-msm8953/vendor_qcom_opensource_wlan_prima, which builds after providing an empty `include/linux/wakelock.h` file in the kernel tree.

WiFi appears to work with it after some modifications (disable BT, try to get it to stop doing assoc and roaming in the driver, provide mac address in `/lib/firmware`, enable with `echo 1 > /dev/wcnss_wlan`, load the module, `rfkill unblock 0`, then start `hostapd` and `wpa_supplicant`, manually configure IP and route, `ssh` works)!

Now I should figure out which of these modifications are actually necessary, and whether a stock driver would be preferable.

MagneFire:
Great work!

For what it's worth these wlan kernel driver might be more suited: https://github.com/Nhannh2791/Linux/tree/master/kernel/private/msm-fossil-cw-modules/wlan since it specifically for `hoki`. However, it's not an official source.

We might be able to use `msm-wifi-enabler` here as well since you needed to execute the same commands as listed here: https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-android/msm-wifi-enabler/msm-wifi-enabler/msm-wifi-enabler.service.

Does `connmanctl` work too (https://wiki.asteroidos.org/index.php/IP_Connection) as opposed to manually executing `rfkill`, `hostapd` and `wpa_supplicant`?

argosphil:
> Does `connmanctl` work too (https://wiki.asteroidos.org/index.php/IP_Connection) as opposed to manually executing `rfkill`, `hostapd` and `wpa_supplicant`?

Yes-ish. I just tried, and got this:
```
Error /net/connman/service/wifi_*_managed_psk: Input/output error
```
But then I tried again, and it worked. Will investigate.

gittrekt:
> Great work!
> 
> For what it's worth these wlan kernel driver might be more suited: https://github.com/Nhannh2791/Linux/tree/master/kernel/private/msm-fossil-cw-modules/wlan since it specifically for `hoki`. However, it's not an official source.
> 
> We might be able to use `msm-wifi-enabler` here as well since you needed to execute the same commands as listed here: https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-android/msm-wifi-enabler/msm-wifi-enabler/msm-wifi-enabler.service.
> 
[...]

PR #342, opened 2026-09-03
hoki: gnss: symlink real vendor HAL service, enable NTP AGPS assistance

### LINKED ISSUE #252: [hoki] GPS support
GPS isn't currently supported on `hoki`.

Getting GPS to work probably involves the following:
- Exploring a WearOS dump to find relevant services for use with GPS/GNSS.
- Starting those services
- Observe the `/system/bin/logcat` because it will report errors which will need to be fixed.

pettcomputers:
## hoki: working GPS, from vendor HAL fix through to a real location fix

Fixes #252.

Full disclosure: I used Claude Code to help me work through this — same as the WiFi PR (#224/its PR), I'm not a kernel developer by trade and this involved a lot of unfamiliar territory (HIDL, D-Bus, ACDB calibration, etc.). Sharing it back here in case it helps someone else.

### What this does

Gets a complete, working GPS location stack on hoki: the vendor GNSS HAL service starts cleanly at boot, `geoclue` reports real position data, and it's been confirmed with an actual moving fix during a drive (13-15 satellites tracked, live-updating coordinates), not just a stationary indoor test.

### Part 1 — vendor GNSS HAL service

The real vendor binary (`android.hardware.gnss@1.0-service-qti`) is present on `/vendor/bin/hw/` and works fine when launched directly. The blocker was that `hal-droid`'s init parses a `.rc` file referencing the service without the `-qti` suffix — a plain filename mismatch, not a missing binary. Fixed with a symlink at first-boot/postinst time:

```
ln -sf android.hardware.gnss@1.0-service-qti /vendor/bin/hw/android.hardware.gnss@1.0-service
```

Confirmed via `journalctl -b` after a cold reboot: clean `starting service 'vendor.gnss_service'` with zero errors, process stays up.

### Part 2 — location-consumer stack (geoclue) — turned out to already be wired up, just broken

I expected to need to add `geoclue` + `geoclue-provider-hybris-binder` myself (that's what Beluga, same msm8937 platform family, uses — plain upstream packages, no device-specific hacks, so it looked like the missing piece). Turns out `recipes-navigation/geoclue/geoclue_%.bbappend` in this tree already has `RDEPENDS on geoclue-provider-hybris-binder` — someone already wired this in at some point, so nothing to add here. The actual blocker was a real bug in the provider itself, not missing wiring.

**Real bug found and fixed** (this part is a separate PR against `meta-asteroid`, not this repo, since `geoclue-provider-hybris.inc` lives there and the bug isn't hoki-specific — see linked PR): `geoclue-provider-hybris-binder`'s init would crash immediately with `qFatal("Failed to register object ...")`. Traced it to `QDBusConnection::lastError()`: `"Using X11 for dbus-daemon autolaunch was disabled at compile time, set your DBUS_SESSION_BUS_ADDRESS instead"`. The binary is installed setuid-root by upstream's own `geoclue-provider-hybris.inc` (`chmod 04755`), and a setuid-root process has real-uid ≠ effective-uid — exactly the condition under which glibc/D-Bus strip environment variables like `DBUS_SESSION_BUS_ADDRESS` as a real, documented security hardening behavior, happening at the dynamic-loader level before `main()` runs. Confirmed the GNSS binder access this binary needs doesn't actually require root on this port — changed the install step from `chmod 04755` to `chmod 0755`. This affects every device using this provider, not just hoki, which is why it's split into its own PR.

```diff
- chmod 04755 ${D}${libexecdir}/geoclue-hybris
+ chmod 0755 ${D}${libexecdir}/geoclue-hybris
```

Confirmed via a real cold reboot: `busctl --user introspect org.freedesktop.Geoclue.Providers.Hybris ...` returns the full expected interface set (`Position`, `Velocity`, `Satellite`), and the service correctly sits dormant until D-Bus auto-activation, then comes up clean on demand — proper `Type=dbus` on-demand behavior, no manual steps.

### Part 3 — actually computing a fix (not just tracking satellites)

Confirmed via a real outdoor test that satellite tracking worked fine on its own (climbed to 6-8 satellites over several minutes) but `GetStatus` never advanced past "Acquiring" and position stayed empty. Turned out to be assistance-data related: this is a WiFi-only device with no cellular modem, and `geoclue-provider-hybris-binder` has two separate AGPS-assistance code paths — a cellular/QOfono one (permanently unavailable here) and a general-network NTP time-injection path that works over any online connection, gated by `NTP_FORCE_INJECT` in `/etc/gps_xtra.ini`. That config file simply didn't exist on this device, so the WiFi-capable assistance path was silently disabled.

**File added:** `/etc/gps_xtra.ini`:
```ini
[ntp]
NTP_FORCE_INJECT=true
```

### Testing

Three real outdoor tests over multiple sessions:
1. First test: no fix, traced to a test-tooling bug (each poll reconnecting to D-Bus, tearing down the GNSS session every ~37s — not a driver issue).
2. Second test, with the tooling fixed: real satellite tracking confirmed (6-8 satellites sustained over 5+ minutes) but no fix — led to the NTP config fix above.
3. Third test, 45 minutes, with the NTP fix in place: full success. Status reached "Available," 13-15 satellites tracked, and real coordinates updated continuously and consistently during an actual drive — a genuine moving GPS track, not a fluke reading.

Open item: no on-watch GPS visualization app currently launches cleanly on this device (a separate booster/IPC issue unrelated to GPS itself, not covered by this PR) — D-Bus polling against the geoclue interface is the practical way to confirm a fix in the meantime.

### DIFF
--- a/meta-hoki/recipes-core/gnss-service-hoki/gnss-service-hoki_1.0.bb
+++ b/meta-hoki/recipes-core/gnss-service-hoki/gnss-service-hoki_1.0.bb
@@ -1,0 +1,28 @@
+DESCRIPTION = "Symlinks the real vendor GNSS HAL binary to the filename AsteroidOS's \
+hal-droid init compat layer actually looks for, and enables NTP-based AGPS \
+assistance data injection over WiFi (this device has no cellular modem for \
+the alternate SUPL assistance path)."
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://gps_xtra.ini;md5=2dfad6f123b25b9c584cc209770743a2"
+PR = "r0"
+
+SRC_URI = "file://gps_xtra.ini"
+S = "${WORKDIR}/sources"
+UNPACKDIR = "${S}"
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+COMPATIBLE_MACHINE = "hoki"
+
+do_install() {
+    install -m 0755 -d ${D}${sysconfdir}
+    install -m 0644 gps_xtra.ini ${D}${sysconfdir}/gps_xtra.ini
+}
+
+pkg_postinst:${PN}() {
+    # hal-droid's init .rc parses a service filename without the "-qti"
+    # suffix, but the real vendor binary is only present as
+    # android.hardware.gnss@1.0-service-qti. Symlink so the service
+    # actually starts. See https://github.com/AsteroidOS/meta-smartwatch/issues/252
+    ln -sf android.hardware.gnss@1.0-service-qti $D/vendor/bin/hw/android.hardware.gnss@1.0-service
+}
+
+FILES:${PN} += "${sysconfdir}/gps_xtra.ini"
--- a/meta-hoki/recipes-core/gnss-service-hoki/gps_xtra.ini
+++ b/meta-hoki/recipes-core/gnss-service-hoki/gps_xtra.ini
@@ -1,0 +1,2 @@
+[ntp]
+NTP_FORCE_INJECT=true


meta-hoki/recipes-core/gnss-service-hoki/gnss-service-hoki_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+    install -m 0755 -d ${D}${sysconfdir}
+    install -m 0644 gps_xtra.ini ${D}${sysconfdir}/gps_xtra.ini
+}
+
+pkg_postinst:${PN}() {
+    # hal-droid's init .rc parses a service filename without the "-qti"
+    # suffix, but the real vendor binary is only present as
+    # android.hardware.gnss@1.0-service-qti. Symlink so the service
+    # actually starts. See https://github.com/AsteroidOS/meta-smartwatch/issues/252
+    ln -sf android.hardware.gnss@1.0-service-qti $D/vendor/bin/hw/android.hardware.gnss@1.0-service
```
MagneFire, line 25: I think the cleaner approach is to define the service in `init.rc` like this:
```
service vendor.gnss_service /vendor/bin/hw/android.hardware.gnss@1.0-service-qti
    class hal
```

The reason why this works is because `init.rc` is loaded first, then the other `.rc` files are loaded in `/usr/libexec/hal-droid/system/etc/init`. Services with the same names are ignored, this means that the qti variant you define in `init.rc` has highest priority.

meta-hoki/recipes-core/gnss-service-hoki/gnss-service-hoki_1.0.bb
```diff
@@ -0,0 +1,28 @@
+DESCRIPTION = "Symlinks the real vendor GNSS HAL binary to the filename AsteroidOS's \
```
MagneFire, line 1: This recipe isn't actually installed by anything, meaning you still need to manually install this via the package manager. Using the method below you also take advantage of not having to do this as `init.rc` is already installed.

meta-hoki/recipes-core/gnss-service-hoki/gps_xtra.ini
```diff
@@ -0,0 +1,2 @@
+[ntp]
+NTP_FORCE_INJECT=true
```
MagneFire, line 2: What happens when there's no network connection? 

Also, this appears to be a `[geoclue-providers-hybris](https://github.com/mer-hybris/geoclue-providers-hybris/tree/master)` feature, would it make sense to make it available for all platforms via a `meta-asteroid` patch (https://github.com/AsteroidOS/meta-asteroid/blob/master/recipes-nemomobile/geoclue-providers/geoclue-provider-hybris.inc)?

MagneFire (commented):
Thanks for your work on this! Some questions popped up.
It mostly consists of getting the things in a consistent state and (hopefully) less specific to `hoki`.

pettcomputers:
Good catches, thanks for taking a look.

On the init.rc approach - that's much cleaner than what I did. I was fighting the "wrong binary name" problem with a symlink because I didn't realize services defined earlier in init.rc take priority over ones with the same name loaded later from the other rc files. I'll switch to defining `vendor.gnss_service` directly in init.rc pointing at the `-qti` binary and drop the symlink and the standalone recipe.

That also takes care of your second point - since init.rc is already part of the base image, there's nothing left that needs a separate install step once the service definition lives there instead of in its own recipe.

On the xtra/network question: NTP_FORCE_INJECT only gates the time-assist path (`injectUtcTime()`/`sendNtpRequest()`), which only fires when `NetworkManager::state() == OnlineState && gpsStarted`. If there's no network, that code path just never runs - it doesn't block or fail the actual GPS acquisition, the receiver still does a normal cold-start fix off satellite data alone, just without the time hint that speeds up ephemeris decode. So worst case with no network is a slower fix, not a broken one.

And yeah, you're right that this isn't really a hoki-specific thing - it's a geoclue-providers-hybris feature, so any device on that provider would benefit from it. I'll move it over to a meta-asteroid patch on geoclue-provider-hybris.inc instead of keeping it hoki-only in this PR, if that sounds right to you.

I'll get an updated PR pushed once I've tested the init.rc change on-device.

MagneFire:
Sounds good, thanks!

One thing about the AI work though: Try to keep it consistent with the codebase, this mostly means reducing the amount of comments placed at changes. If people really want to know more about why a change was made they can look at the relevant git commit message. This also means that the git commit message should really contain *why* a change was done. AI is also really good at going into excruciating details, which only makes things more confusing.

This is just a heads up, you're doing great work!

